PR #16, opened 2015-01-07
Graceful cluster restart on domain exception

briangough:
Handle exceptions gracefully using cluster to restart processes, avoid memory leak from old domains

Uses the 'recluster' module.


jpallen:
LGTM @henryoswald ?

PR #19, opened 2015-11-20
Add a PersistorManager library using the aws-sdk library

xtrochu:
This changes add a new PersistorManager implementation that utilise the aws-sdk library.

This helps when the Filestore server is hosted on AWS and is expecting to get access to the S3 store  without credentials via an IAM Role.


henryoswald:
Think this should be fine. Looks like some of the files are using 2 spaces rather than 1 tab, could you change them?

xtrochu:
Hi.

I fixed the indentation on the two new files.

PR #24, opened 2017-06-21
If optipng is sigkilled by timeout, then ignore that error

ShaneKilkelly:


app/coffee/ImageOptimiser.coffee
```diff
@@ -11,10 +11,12 @@ module.exports =
 			timeout: 30 * 1000
 			killSignal: "SIGKILL"
 		exec args, opts,(err, stdout, stderr)->
+			if err? and err?.signal == 'SIGKILL'
+				logger.warn {err: err, stderr: stderr, localPath: localPath}, "optimiser timeout reached"
+				err = null
 			if err?
```
henryoswald: this if err? will still be hit, make it an else if?

ShaneKilkelly: Refactored

henryoswald: it will still log out the problem twice, not big deal but is that intentional?

app/coffee/ImageOptimiser.coffee
```diff
@@ -12,9 +12,12 @@ module.exports =
 			killSignal: "SIGKILL"
 		exec args, opts,(err, stdout, stderr)->
 			if err?
-				logger.err err:err, stderr:stderr, localPath:localPath, "something went wrong converting compressPng"
+				if err.signal == 'SIGKILL'
```
henryoswald: why do nested ifs, what about 

	if err? and err.signal == 'SIGKILL'
		...
	else if err?
		...
	else

ShaneKilkelly: Ah, yes, misunderstood the first comment, will refactor.

PR #28, opened 2018-07-04
Add endpoint for arbitrary bucket fetch

mmazour:
Add `/bucket/:bucket/key/*`, which fetches the file from the given bucket at the given path. Uses auth stored at `settings.filestore.s3.{{bucketName}}` if present, and otherwise default auth.

This is to support the use case where:
- v1 makes a CLSI request to v2, 
    and
- request contains file URLs stored on S3, 
    and
- they are not in the default bucket that v2 would use for its own files, 
    and
- CLSI is in a contained environment and cannot reach S3 except through filestore.

Child of https://github.com/overleaf/sharelatex/issues/664
Child of https://github.com/overleaf/sharelatex/issues/507

### DIFF
--- a/app/coffee/BucketController.coffee
+++ b/app/coffee/BucketController.coffee
@@ -7,8 +7,8 @@ Errors = require('./Errors')
 module.exports = BucketController =
 
 	getFile: (req, res)->
-		{bucket} = req
-		key = req[0]
+		{bucket} = req.params
+		key = req.params[0]
 		{format, style} = req.query
 		credentials = settings.filestore.s3&[bucket]
 		options = {
--- a/app/coffee/BucketController.coffee
+++ b/app/coffee/BucketController.coffee
@@ -1,13 +1,8 @@
-PersistorManager = require("./PersistorManager")
 settings = require("settings-sharelatex")
 logger = require("logger-sharelatex")
 FileHandler = require("./FileHandler")
 metrics = require("metrics-sharelatex")
-parseRange = require('range-parser')
 Errors = require('./Errors')
-
-oneDayInSeconds = 60 * 60 * 24
-maxSizeInBytes = 1024 * 1024 * 1024 # 1GB
 
 module.exports = BucketController =
 
@@ -21,7 +16,7 @@ module.exports = BucketController =
 			bucket: bucket,
 			credentials: credentials
 		}
-		metrics.inc "getFile"
+		metrics.inc "#{bucket}.getFile"
 		logger.log key:key, bucket:bucket, "receiving request to get file from bucket"
 		FileHandler.getFile bucket, key, options, (err, fileStream)->
 			if err?
@@ -33,4 +28,3 @@ module.exports = BucketController =
 			else
 				logger.log key:key, bucket:bucket, format:format, style:style, "sending bucket file to response"
 				fileStream.pipe res
-
--- a/app/coffee/BucketController.coffee
+++ b/app/coffee/BucketController.coffee
@@ -9,7 +9,6 @@ module.exports = BucketController =
 	getFile: (req, res)->
 		{bucket} = req.params
 		key = req.params[0]
-		{format, style} = req.query
 		credentials = settings.filestore.s3[bucket]
 		options = {
 			key: key,
@@ -20,11 +19,11 @@ module.exports = BucketController =
 		logger.log key:key, bucket:bucket, "receiving request to get file from bucket"
 		FileHandler.getFile bucket, key, options, (err, fileStream)->
 			if err?
-				logger.err err:err, key:key, bucket:bucket, format:format, style:style, "problem getting file from bucket"
+				logger.err err:err, key:key, bucket:bucket, "problem getting file from bucket"
 				if err instanceof Errors.NotFoundError
 					return res.send 404
 				else
 					return res.send 500
 			else
-				logger.log key:key, bucket:bucket, format:format, style:style, "sending bucket file to response"
+				logger.log key:key, bucket:bucket, "sending bucket file to response"
 				fileStream.pipe res
--- a/app/coffee/BucketController.coffee
+++ b/app/coffee/BucketController.coffee
@@ -9,7 +9,7 @@ module.exports = BucketController =
 	getFile: (req, res)->
 		{bucket} = req.params
 		key = req.params[0]
-		credentials = settings.filestore.s3[bucket]
+		credentials = settings.filestore.s3?[bucket]
 		options = {
 			key: key,
 			bucket: bucket,
--- a/app/coffee/BucketController.coffee
+++ b/app/coffee/BucketController.coffee
@@ -10,7 +10,7 @@ module.exports = BucketController =
 		{bucket} = req.params
 		key = req.params[0]
 		{format, style} = req.query
-		credentials = settings.filestore.s3&[bucket]
+		credentials = settings.filestore.s3[bucket]
 		options = {
 			key: key,
 			bucket: bucket,
--- a/app/coffee/BucketController.coffee
+++ b/app/coffee/BucketController.coffee
@@ -1,0 +1,36 @@
+PersistorManager = require("./PersistorManager")
+settings = require("settings-sharelatex")
+logger = require("logger-sharelatex")
+FileHandler = require("./FileHandler")
+metrics = require("metrics-sharelatex")
+parseRange = require('range-parser')
+Errors = require('./Errors')
+
+oneDayInSeconds = 60 * 60 * 24
+maxSizeInBytes = 1024 * 1024 * 1024 # 1GB
+
+module.exports = BucketController =
+
+	getFile: (req, res)->
+		{bucket} = req
+		key = req[0]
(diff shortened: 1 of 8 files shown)


app/coffee/BucketController.coffee
```diff
@@ -0,0 +1,36 @@
...
+oneDayInSeconds = 60 * 60 * 24
+maxSizeInBytes = 1024 * 1024 * 1024 # 1GB
+
+module.exports = BucketController =
+
+	getFile: (req, res)->
+		{bucket} = req
+		key = req[0]
+		{format, style} = req.query
+		credentials = settings.filestore.s3&[bucket]
```
henryoswald: Is the & meant to be there? I've not seen it before. or are you trying todo a null check?

app/coffee/BucketController.coffee
```diff
@@ -0,0 +1,36 @@
...
+parseRange = require('range-parser')
+Errors = require('./Errors')
+
+oneDayInSeconds = 60 * 60 * 24
+maxSizeInBytes = 1024 * 1024 * 1024 # 1GB
+
+module.exports = BucketController =
+
+	getFile: (req, res)->
+		{bucket} = req
```
henryoswald: is bucket not on `req.params` rather than just `req`? Might well but but its new to me

jpallen:
> CLSI is in a contained environment and cannot reach S3 except through filestore.

Where is this logic implemented? Would it be easier to extend this logic (wherever it is), to also allow access to s3 (possibly just some whitelisted buckets)?

app/coffee/BucketController.coffee
```diff
@@ -0,0 +1,36 @@
+PersistorManager = require("./PersistorManager")
+settings = require("settings-sharelatex")
+logger = require("logger-sharelatex")
+FileHandler = require("./FileHandler")
+metrics = require("metrics-sharelatex")
+parseRange = require('range-parser')
+Errors = require('./Errors')
+
+oneDayInSeconds = 60 * 60 * 24
+maxSizeInBytes = 1024 * 1024 * 1024 # 1GB
```
henryoswald: I don't think these var are used?

app/coffee/BucketController.coffee
```diff
@@ -0,0 +1,36 @@
+PersistorManager = require("./PersistorManager")
+settings = require("settings-sharelatex")
+logger = require("logger-sharelatex")
+FileHandler = require("./FileHandler")
+metrics = require("metrics-sharelatex")
+parseRange = require('range-parser')
+Errors = require('./Errors')
```
henryoswald: PersistorManager & parseRange are not used

app/coffee/BucketController.coffee
```diff
@@ -0,0 +1,36 @@
...
+		{bucket} = req
+		key = req[0]
+		{format, style} = req.query
+		credentials = settings.filestore.s3&[bucket]
+		options = {
+			key: key,
+			bucket: bucket,
+			credentials: credentials
+		}
+		metrics.inc "getFile"
```
henryoswald: shall we put the bucket into this metric so we can see the different loads. i.e. `metrics.inc "#{bucket}.getFile"`

app/coffee/BucketController.coffee
mmazour: I'm trying to do a null check.

app/coffee/BucketController.coffee
mmazour: I was following the pattern of https://github.com/sharelatex/filestore-sharelatex/blob/master/app/coffee/FileController.coffee#L15

app/coffee/BucketController.coffee
mmazour: Yes, sorry, I'll prune out the unused stuff.

app/coffee/BucketController.coffee
henryoswald: `settings.filestore.s3?[bucket]` would null check if s3 exists

app/coffee/BucketController.coffee
henryoswald: I think bucket gets injected with `keyBuilder.userFileKey` https://github.com/sharelatex/filestore-sharelatex/blob/master/app.coffee#L70

henryoswald:
@jpallen discussion is happening in the parent ticket overleaf/sharelatex#664

mmazour:
> >CLSI is in a contained environment and cannot reach S3 except through filestore.
> 
> Where is this logic implemented? Would it be easier to extend this logic (wherever it is), to also allow access to s3 (possibly just some whitelisted buckets)?

@jpallen this is from https://github.com/overleaf/sharelatex/issues/664#issuecomment-401810172

henryoswald:
@mmazour if you put some credentials into the settings.defaults.coffee file it should be possible to test this with a quick curl

app/coffee/BucketController.coffee
mmazour: What I want is that if s3 doesn't exist, or if s3[bucket] doesn't exist, you get a null value and don't throw an error.

app/coffee/BucketController.coffee
mmazour: Sorry, I see these things and I just assume it's some magic in the package that I didn't know about before. Fixing.

mmazour:
@henryoswald I'll give it a go.

mmazour:
Back to WIP until I've made that work and we've decided whether we're doing this change.

mmazour:
Another couple of changes, and successfully tested with a modified settings file.
```
$ http localhost:3009/bucket/writelatex-conversions-dev/key/clsi/overleaftest/Overleaf_5b362c60497c794809c3ef4e_v3
HTTP/1.1 200 OK
Connection: keep-alive
Date: Wed, 04 Jul 2018 15:38:56 GMT
Transfer-Encoding: chunked
X-Powered-By: Express

+-----------------------------------------+
| NOTE: binary data not shown in terminal |
+-----------------------------------------+
```
Back to Review.

app/coffee/BucketController.coffee
```diff
@@ -0,0 +1,30 @@
...
+metrics = require("metrics-sharelatex")
+Errors = require('./Errors')
+
+module.exports = BucketController =
+
+	getFile: (req, res)->
+		{bucket} = req.params
+		key = req.params[0]
+		{format, style} = req.query
+		credentials = settings.filestore.s3[bucket]
```
henryoswald: It is worth checking that the credentials exist or an error will be thrown further down the line.

app/coffee/BucketController.coffee
```diff
@@ -0,0 +1,30 @@
...
+FileHandler = require("./FileHandler")
+metrics = require("metrics-sharelatex")
+Errors = require('./Errors')
+
+module.exports = BucketController =
+
+	getFile: (req, res)->
+		{bucket} = req.params
+		key = req.params[0]
+		{format, style} = req.query
```
henryoswald: format, style are not used (also logged further down)

app/coffee/BucketController.coffee
henryoswald: Can we add the settings to  settings.defaults.coffee with corresponding env variables to set them. The env might be a bit messy, everything in the k8 world is an env variable so it might have to be a JSON string we pass through then settings does a JSON.parse

mmazour:
@henryoswald I think I've addressed your review comments. Have another look?

Also LMK if you like having the `SettingsTests` in there, or if having settings tests is weird and you'd rather not.

app/coffee/BucketController.coffee
```diff
@@ -0,0 +1,29 @@
...
+FileHandler = require("./FileHandler")
+metrics = require("metrics-sharelatex")
+Errors = require('./Errors')
+
+module.exports = BucketController =
+
+	getFile: (req, res)->
+		{bucket} = req.params
+		key = req.params[0]
+		credentials = settings.filestore.s3?[bucket]
```
henryoswald: Can we check that these credentials exist before passing them through to the rest of the stack.

mmazour: My intention was that if they didn't exist, it would fall back to the default s3 credentials. That logic is in `S3PersistorManager`:
```
	getFileStream: (bucketName, key, opts, callback = (err, res)->)->
		opts = opts || {}
		headers = {}
		if opts.start? and opts.end?
			headers['Range'] = "bytes=#{opts.start}-#{opts.end}"
		callback = _.once callback
		logger.log bucketName:bucketName, key:key, "getting file from s3"
		s3Client = knox.createClient
			key: opts.credentials?.auth_key || settings.filestore.s3.key
			secret: opts.credentials?.auth_secret || settings.filestore.s3.secret
			bucket: bucketName
```

henryoswald: Ok, I didn't like that before but actually it is fair enough as it gives us the ability to use this endpoint for other things if we needed.

henryoswald:
I just need to run some curl's then will aprove

mmazour:
Great, thanks. Squashed down and ready for merge next week.

PR #29, opened 2018-07-09
Decaffeinate service

afcapel:


natestemen:
going to close. we _are_ going to decaffeinate, but i think we are going to break it down into a few more steps to preserve git history and such.

PR #37, opened 2018-11-09
reduce logging of not found errors

briangough:
We are logging huge numbers of `Not Found` errors from loading templates which either don't have thumbnails (the creation of the thumbnail failed for some reason) or the template pdf wasn't uploaded. We can at least suppress these from sentry by not logging them as errors.


briangough:
@ShaneKilkelly I'm going to merge this and deploy it's a small change.

PR #43, opened 2019-01-07
Disable exec with settings flag

henryoswald:
We can't really trust the convert commands in filestore on k8 due to lack of firejail sandboxing, this flag hard disables there functionality until we have a better system available. I've not checked on user facing impact of previewing `eps` and `pdf` image in the editor yet, this PR is just for security reasons.

Child of https://github.com/overleaf/sharelatex/issues/552


briangough:
Are you sure you want to make this pull request against `master` - should it be against `ho-docker`?

henryoswald:
No, you are correct!

PR #44, opened 2019-01-09
return a 404 error (instead of a 500) when copying a missing file

briangough:
This returns a 404 error (instead of a 500) when copying a missing file.  We have many projects lurking with missing files and we need a way to detect this in web if we want to handle it gracefully.

connects to https://github.com/overleaf/sharelatex/issues/1225


jpallen:
(NB I've added filestore to waffle now, so this should show up)

PR #46, opened 2019-01-22
Decaffeinate filestore

natestemen:
fixes https://github.com/sharelatex/filestore-sharelatex/issues/45


natestemen:
I think we are going to finish the google migration before moving this forward, and i think this is sufficiently outdated to start fresh so I'm going to close this out for now.

PR #47, opened 2019-02-05
Move to docker build process

henryoswald:
A lot of this code has been running as part of filestore-readonly in prod for a while.

This pr builds a docker image and tar.gz for linode & gke deployments.

Sorry about the number of commits, squashing them got messy so I gave up.


briangough:
I was able to get this running with
```
$ cd Code/Overleaf/overleaf
$ ./bin/pull && (cd filestore && git fetch && git checkout ho-docker)
$ docker-compose build
$ ./bin/install
```

To run the tests

```
$ (cd filestore && make test_unit) # OK
$ (cd filestore && make test_acceptance) 

  10 passing (165ms)
  2 failing

  1) Filestore
       with a pdf file
         getting the preview image
           should respond with image data:

      Uncaught AssertionError: expected 500 to equal 200
      + expected - actual

      -500
      +200
      
      at Request._callback (test/acceptance/js/SendingFileTest.js:195:46)
      at Request.self.callback (node_modules/request/main.js:127:22)
      at Request.<anonymous> (node_modules/request/main.js:767:16)
      at IncomingMessage.<anonymous> (node_modules/request/main.js:729:14)
      at endReadableNT (_stream_readable.js:974:12)
      at _combinedTickCallback (internal/process/next_tick.js:74:11)
      at process._tickDomainCallback (internal/process/next_tick.js:122:9)
  

  2) Filestore
       with a pdf file
         warming the cache
           should respond with only an 'OK':

      Uncaught AssertionError: expected 500 to equal 200
      + expected - actual

      -500
      +200
      
      at Request._callback (test/acceptance/js/SendingFileTest.js:219:46)
      at Request.self.callback (node_modules/request/main.js:127:22)
      at Request.<anonymous> (node_modules/request/main.js:767:16)
      at IncomingMessage.<anonymous> (node_modules/request/main.js:729:14)
      at endReadableNT (_stream_readable.js:974:12)
      at _combinedTickCallback (internal/process/next_tick.js:74:11)
      at process._tickDomainCallback (internal/process/next_tick.js:122:9)
  
```

If I delete the whole repo more of the acceptance tests fail because the directories created from `install_deps.sh`

```
mkdir /app/user_files/ /app/uploads/ /app/template_files/
chown -R node:node /app/user_files
chown -R node:node /app/uploads
chown -R node:node /app/template_files
```
are not being created outside the container.  This may be a Linux specific issue.

Dockerfile
```diff
@@ -1,4 +1,23 @@
...
+
+COPY . /app
+
+
+RUN npm run compile:all
+
 FROM node:6.9.5
 
-# we also need imagemagick but it is already in the node docker image
-RUN apt-get update && apt-get install -y --no-install-recommends ghostscript optipng
```
briangough: Any reason we are switching from installing `ghostscript` from a package to compiling from source?  If we used an appropriate base image we could get security updates for ghostscript automatically.

install_deps.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+apt-get update
+
+apt-get install vim imagemagick optipng --yes
+
+wget -q https://s3.amazonaws.com/sl-public-dev-assets/ghostscript-9.15.tar.gz -O /tmp/ghostscript-9.15.tar.gz
```
briangough: Any reason to use version 9.15? The latest is 9.26.

henryoswald:
I've fixed the failing tests, I hadn't flipped a featured flag locally, it was only on the ci compose file.

Are `mkdir /app/user_files/ /app/uploads/ /app/template_files/` not there with the git repo?

briangough:
Thanks, that fixed it (also doing a `git checkout` to get those directories).

On the docker version there is `node:6-stretch` which would have security updates and avoid needing to compile our own ghostscript.  I tried switching to that - it seemed to work, and pulled in the latest Debian ghostscript (9.26a~dfsg-0+deb9u1).  What do you think @mans0954 ?

henryoswald:
We are not actually going to be running ghostscript in this container as we didn't manage to secure it so there is an argument we remove it here. Is stretch just a more complete debian?

briangough:
Doh, should have realised that. If we're not using ghostscript in production then this is not worth worrying about now.

`stretch` is the current Debian LTS.  The node-6.9.5 image was on the older Debian release (`jessie`) which is not getting security updates any more.

henryoswald:
Thanks. FYI once we get everything moved to k8 we will be aggressively moving everything to node 10.  Having it all dockerised should make is less painful.

PR #49, opened 2019-05-02
[misc] add support for node 10+

das7pad:
This PR adds support for node version 10 and above.

Attaching a `readable` listener to a `Readable` stream causes the stream to hang on node version 10 [1].

REF https://github.com/sharelatex/filestore-sharelatex/pull/40

---
[1] See last sentence in this paragraph: https://nodejs.org/api/stream.html#stream_two_reading_modes


ShaneKilkelly:
Hi @das7pad ! Thanks for this PR. We'll get around to merging it soon, hopefully :)

gh2k:
Thank you for the fix. :-)

PR #51, opened 2019-05-02
[tests] cleanup resources

das7pad:
The acceptance tests leave plenty of files behind them.

Namely the test fixtures for reading / writing in `/tmp`, the files pushed during the setup of each test case in `user_files` and `uploads`, the copied project file in `user_files` and the preview images in the `user_files`.

The application is started but never stopped. `npm run` would kill the server, other (ide-)tools may leave it running without explicit configuration. This slows down local development as the app has to be killed manually after each test run.

This PR adds the cleanup of the mentioned files grouped per section and stops the application afterwards.

I suggest a diff view without white space changes, the diff in gh is too verbose.

### DIFF
--- a/test/acceptance/coffee/FilestoreApp.coffee
+++ b/test/acceptance/coffee/FilestoreApp.coffee
@@ -2,10 +2,12 @@ app = require('../../../app')
 require("logger-sharelatex").logger.level("info")
 logger = require("logger-sharelatex")
 Settings = require("settings-sharelatex")
+Metrics = require("metrics-sharelatex")
 
 module.exports =
 	running: false
 	initing: false
+	server: null
 	callbacks: []
 	ensureRunning: (callback = (error) ->) ->
 		if @running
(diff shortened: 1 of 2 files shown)


gh2k:
Are you running tests via the Makefile+Docker? The idea is that we tear down and recreate the containers after each test run, so anything left behind is nuked that way. I think this is a better way of running acceptance tests anyway, as you can always be sure you're not accidentally depending on an artifact that you've left lying around.

das7pad:
I am running tests via the mocha integration of my IDE (Intellij IDEA) and via docker-compose in CI. Omitting the docker overhead is much faster and easier to invoke for me [0].

Upon a review of this PR, I would drop most of the changes [1], but retain the cleanup of files in `/tmp`. 
If you agree that files outside the project directory should be cleaned up, I will force push a patch to do exactly that and drop the other changes. Otherwise you can close this PR.


> you can always be sure you're not accidentally depending on an artifact that you've left lying around

This can be seen from a different point of view too: not cleaning up properly could pollute the fs silently and propagate to production later. We could use the acceptance tests to verify that all routes cleanup their (temporary) files behind them. This check may be placed in a global `after` hook.
But this is not was this PR was about initially and a future issue/PR may cover this instead. I do not have time to bring this idea forward at the moment.

---
[0]: docker: my run user is explicitly not added to the docker user group for security reasons; rootless docker is not yet stable. general: the IDEA integration is awesome for partial runs of the test suites - e.g. select a single case/suite to run or rerun just failed cases, open the debugger on a subset of tests. This in turn requires proper encapsulation of tests.
[1]: The stopping of the server can be replaced with the mocha flag `--exit` on my end. The files generated by the app are already scoped in their `XXX_files` directories. Leaving no files behind would be great, but the way I implemented this _in May_ is too aware of app details - in particular the composing of the file paths. Using the DEL routes may be a cleaner way to delete these files.

gh2k:
Thanks. I agree running outside Docker is faster, but we try to run things consistently across mac/pc, so generally stick with doing it that way.

I'm going to close this for now as I think it doesn't suit our internal workflow, but thanks for taking the time to work on it.

PR #53, opened 2019-06-18
Use AWS SDK for getFileStream()

emcsween:
### Description

The AWS SDK has a retry strategy to deal with rate limiting or transient
unavailability of S3. We hope it will reduce our error rates.

#### Related Issues / PRs

* https://github.com/overleaf/issues/issues/1793

#### Potential Impact

This piece of code is called from the following endpoints:

* GET /project/:project_id/file/:file_id
* GET /template/:template_id/v/:version/:format
* GET /project/:project_id/public/:public_file_id
* GET /bucket/:bucket/:key/*

#### Manual Testing Performed

- [ ] Upload a file in a project and preview it
- [ ] Get a file directly from Filestore
- [ ] Try to get a non-existent file from Filestore and expect a 404

#### Metrics and Monitoring

See if that stops us from failing on 503s from S3 (https://sentry.io/organizations/overleaf/issues/682378420/?project=60782)

### DIFF
--- a/test/unit/coffee/S3PersistorManagerTests.coffee
+++ b/test/unit/coffee/S3PersistorManagerTests.coffee
@@ -17,18 +17,27 @@ describe "S3PersistorManagerTests", ->
 					key: "this_key"
 				stores:
 					user_files:"sl_user_files"
-		@stubbedKnoxClient =
+		@knoxClient =
 			putFile:sinon.stub()
 			copyFile:sinon.stub()
 			list: sinon.stub()
 			deleteMultiple: sinon.stub()
 			get: sinon.stub()
 		@knox =
-			createClient: sinon.stub().returns(@stubbedKnoxClient)
-		@stubbedS3Client =
+			createClient: sinon.stub().returns(@knoxClient)
+		@s3EventHandlers = {}
+		@s3Request =
+			on: sinon.stub().callsFake (event, callback) =>
+				@s3EventHandlers[event] = callback
+			send: sinon.stub()
+		@s3Response =
+			httpResponse:
+				getUnbufferedStream: sinon.stub()
+		@s3Client =
 			copyObject: sinon.stub()
 			headObject: sinon.stub()
-		@awsS3 = sinon.stub().returns @stubbedS3Client
+			getObject: sinon.stub().returns(@s3Request)
+		@awsS3 = sinon.stub().returns(@s3Client)
 		@LocalFileWriter =
 			writeStream: sinon.stub()
 			deleteFile: sinon.stub()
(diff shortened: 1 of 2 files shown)


test/unit/coffee/S3PersistorManagerTests.coffee
```diff
@@ -50,121 +59,95 @@ describe "S3PersistorManagerTests", ->
...
+					expect(@s3Client.getObject.lastCall.args).to.deep.equal([{
+						Bucket: @bucketName
+						Key: @key
+						Range: "bytes=#{start}-#{end}"
+					}])
+					expect(stream).to.equal(@expectedStream)
+					done()
+
+		describe "errors", ->
+			it "returns NotFoundError if the file doesn't exist", (done) ->
```
das7pad: The part `if the file doen't exist`  should rather stay into a `describe` block.

In case of a failure, the condition and expected code behavior are separated.

test/unit/coffee/S3PersistorManagerTests.coffee
```diff
@@ -50,121 +59,95 @@ describe "S3PersistorManagerTests", ->
...
+				@s3Request.send.callsFake () =>
+					@s3EventHandlers.httpHeaders(404, {}, @s3Response, "Not found")
+				@S3PersistorManager.getFileStream @bucketName, @key, {}, (err, stream) =>
+					expect(err).to.be.instanceof(@Errors.NotFoundError)
+					errMsg = @Errors.NotFoundError.lastCall.args[0]
+					expect(errMsg).to.match(new RegExp(".*#{@bucketName}.*"))
```
das7pad: This is not was the `it` message requested. It is an additional requirement for our error: we want it to include additional context.

A test failure would hint from its message that the error is not a `NotFoundError`, while the actual error could be the missing context in the error message.

test/unit/coffee/S3PersistorManagerTests.coffee
```diff
@@ -50,121 +59,95 @@ describe "S3PersistorManagerTests", ->
...
+				@expectedStream = { expectedStream: true }
+				@s3Request.send.callsFake () =>
+					@s3EventHandlers.httpHeaders(200, {}, @s3Response, "OK")
+				@s3Response.httpResponse.getUnbufferedStream.returns(@expectedStream)
+
+			it "returns a stream", (done) ->
+				@S3PersistorManager.getFileStream @bucketName, @key, @opts, (err, stream) =>
+					if err?
+						return done(err)
+					expect(@awsS3.lastCall.args).to.deep.equal([{
```
das7pad: This is a test on the credentials, while a test failure would hint that something other that a stream was returned.

emcsween:
Thanks @das7pad. I split the tests as suggested. Let me know what you think.

test/unit/coffee/S3PersistorManagerTests.coffee
```diff
@@ -117,29 +129,38 @@ describe "S3PersistorManagerTests", ->
...
+					@S3PersistorManager.getFileStream @bucketName, @key, {}, (err, stream) =>
+						expect(err).to.be.instanceof(Error)
+						done()
+
+			describe "when the S3 request errors out before receiving HTTP headers", ->
+				beforeEach ->
+					@s3Request.send.callsFake () =>
+						@s3EventHandlers.error(new Error("connection failed"))
+
+				it "returns an error if the request errors out before receiving HTTP headers", (done) ->
```
das7pad: This test has the condition in the title.

jdleesmiller:
Thanks for the feedback @das7pad .

I think there can be tradeoffs here, but in this case I agree with the change. I have seen the 'one assertion per test' style cause test performance problems when the test setup is expensive, but here it looks like we're stubbing most things out, so that's not an issue, and the added clarity is probably a win.

das7pad:
You are welcome.

For assertion-only tests the `before` setup might be better suited then `beforeEach`. It might be worth to check this for all projects that run slow unit/acceptance tests.

My Jenkins estimated 5 secs for the unit tests in this project - that's from 37 runs. I do not think we should spend time to optimize this.

But for the web-project it might be worth it - the test runs there average each between one and two minutes - depending on the number of modules ;).

PR #58, opened 2019-12-05
Add metrics for S3 ingress and egress, and improve acceptance tests

gh2k:
### Description

Primarily, this adds `s3.ingress` and `s3.egress` metrics to Filestore.

In order to test this, I wanted to write an acceptance test that verifies the metric. This involved:
* Adding a fake S3 container to the docker test environment
* Configuring the test environment to use S3 instead of the filesystem
* Setting `USE_PROM_METRICS` so that we can scrape the `/metrics` endpoint

### Review

Removed the unused `AWSSDKPersistorManager`. It's confusing having two AWS persistors.

Renamed a local `request` variable to `s3request`. The former was overwriting the global `request` object in a way that broke other methods in the file. This only became apparent when S3 was being used for the accpetance tests. Specifically, the methods were `deleteFile` and `checkIfFileExists`. I don't think we have anything currently running in production that does this.

#### Manual Testing Performed

- [ ] Upload a file, check metrics
- [ ] Download a file, check metrics

### DIFF
--- a/app/coffee/S3PersistorManager.coffee
+++ b/app/coffee/S3PersistorManager.coffee
@@ -132,6 +132,9 @@ module.exports =
 				logger.log({bucketName: bucketName, key: key }, "error getting file from s3: #{statusCode}")
 				return callback(new Error("Got non-200 response from S3: #{statusCode} #{statusMessage}"), null)
 			stream = response.httpResponse.createUnbufferedStream()
+			stream.on 'data', (data) ->
+				metrics.count 's3.ingress', data.byteLength
+
 			callback(null, stream)
 
 		s3Request.on 'error', (err) =>
--- a/app/coffee/S3PersistorManager.coffee
+++ b/app/coffee/S3PersistorManager.coffee
@@ -43,8 +43,9 @@ getS3Options = (credentials) ->
 			secretAccessKey: credentials.auth_secret
 
 	if settings.filestore.s3.endpoint
+		endpoint = URL.parse(settings.filestore.s3.endpoint)
 		options.endpoint = settings.filestore.s3.endpoint
-		options.sslEnabled = false
+		options.sslEnabled = endpoint.protocol == 'https'
 
 	return options
 
--- a/app/coffee/S3PersistorManager.coffee
+++ b/app/coffee/S3PersistorManager.coffee
@@ -16,10 +16,15 @@ LocalFileWriter = require("./LocalFileWriter")
 Errors = require("./Errors")
 _ = require("underscore")
 awsS3 = require "aws-sdk/clients/s3"
+URL = require('url')
 
 thirtySeconds = 30 * 1000
 
 buildDefaultOptions = (bucketName, method, key)->
+	if settings.filestore.s3.endpoint
+		endpoint = "#{settings.filestore.s3.endpoint}/#{bucketName}"
+	else
+		endpoint = "https://#{bucketName}.s3.amazonaws.com"
 	return {
 			aws:
 				key: settings.filestore.s3.key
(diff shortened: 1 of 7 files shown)


gh2k:
I should point out that I think it's slightly unsatisfactory having three mechanisms to talk to S3 in the one file, but this is where we are. :wink:

app/coffee/S3PersistorManager.coffee
```diff
@@ -5,51 +5,74 @@ https.globalAgent.maxSockets = 300
...
-})
+getS3Options = (credentials) ->
+	options =
+		credentials:
+			accessKeyId: credentials.auth_key
+			secretAccessKey: credentials.auth_secret
+
+	if settings.filestore.s3.endpoint
+		options.endpoint = settings.filestore.s3.endpoint
+		options.sslEnabled = false
```
briangough: If ssl is disabled we should make that explicit in the settings, otherwise it would be to easy to think is ssl is being used when it isn't.  I suggest adding a setting for this which defaults to true if not present.

briangough: That was an old version of my comment.  Actually I think you could test on the URL protocol and only disable ssl when `proto` is 'http'.

das7pad:
Just FYI, I did the deprecation [the other way around](https://github.com/das7pad/filestore-sharelatex/commit/fddc7caa3e0ae9b7e44da1e0692485028ba63aae) and continue to use the `AWSSDKPersistorManager`.
Its source is much cleaner and uses the official sdk as the one and only method to interact with s3, which is probably easier for moving forward too.

Acceptance tests passed (with minor tweaks, probably just https://github.com/das7pad/filestore-sharelatex/commit/2957bbceb8adbc5b95cd2a6686117b064d363c33, https://github.com/das7pad/filestore-sharelatex/commit/25a44af8877306f2bec0a0acb85ac0f37fe59dab, https://github.com/das7pad/filestore-sharelatex/commit/929e7fc6d4639906007792547e8cadc602219dda), so I am not sure what the problem with this backend is.

_I am using a [minio cluster](https://github.com/minio/minio) as s3 backend in my infra and for the acceptance tests a single minio node https://github.com/das7pad/filestore-sharelatex/commit/7e0780436cf16a4cd53f78d97eece722e1f4d027 https://github.com/das7pad/filestore-sharelatex/commit/b87716b69ca769a3fd2f5cd8172da24ddb1208ae._

emcsween:
Apart from Brian's comment, this LGTM.

I think it would be nice to merge AWSSDKPersistorManager and S3PersistorManager, basically using the AWS S3 client everywhere, and making sure that we don't lose functionality that was added only in S3PersistorManager. This could be a separate ticket and PR, but if we delete AWSSDKPersistorManager in this PR, we should remind ourselves in the ticket that this file existed at some point.

briangough:
Right I had forgotten the history of the two S3 modules.  The plan was to migrate S3PersistorManager to the AWS SDK, copying the methods from AWSSDKPersistorManager (started here in 2018 https://github.com/overleaf/filestore/pull/39) and when confident that everything was working ok in production then we delete one of them and PersistorManager would load the same backend for s3 and aws-sdk.  I think it's better to remove that deletion commit https://github.com/overleaf/filestore/pull/58/commits/b52258caa2327399ace506b6a99bdcdcf0b76e85 from this PR for now as it's a separate issue.

gh2k:
Thanks, that makes sense. It's (both the S3 implementation, and the fact that we have two aws implementations) a very odd collection of bits & pieces that definitely needs cleaning up.

I'm not clear on what the difference is between the `aws-sdk` and `s3` implementations is in functionality. Thinking about it, as this is an OSS repo there's also an argument that someone could *in theory* have configured their stuff to use `aws-sdk` via settings, so putting it back is indeed a good idea.

It's confusing though. I spent a while trying to change `AWSSDK*` and not understanding why nothing was happening. :man_shrugging:

das7pad:
`connect ECONNREFUSED 172.19.0.3:9090` this seems to be [the fakeS3 container](https://github.com/overleaf/filestore/pull/58/files#diff-d167ab54631a74ff3051701febe45dc0)?

das7pad:
The ci docker-compose config for the s3 container is invalid. Its using a command argument instead of an environment variable for the `initialBuckets`, see the non ci docker-compose config.

docker-compose.ci.yml
```diff
@@ -46,3 +54,7 @@ services:
 
   mongo:
     image: mongo:3.4
+
+  fakes3:
+    image: adobe/s3mock
+    command: --initialBuckets=fake_user_files,fake_template_files,fake_public_files
```
das7pad:
```suggestion
    environment:
      - initialBuckets=fake_user_files,fake_template_files,fake_public_files
```

gh2k:
> The ci docker-compose config for the s3 container is invalid. Its using a command argument instead of an environment variable for the `initialBuckets`, see the non ci docker-compose config.

You are indeed correct. Thanks :-) - Apologies for not using your change, I'd made mine at the same time.

gh2k:
Fixed the tests. I'm not sure why it was necessary to put an explicit is-S3-alive check in the acceptance tests for the cloud builder tests to pass when using a docker-compose healthcheck worked for Jenkins, but here we are.

PR #60, opened 2019-12-16
Upgrade to Node 10 and decaffeinate

gh2k:
### Description

This does a full decaf of the project, and upgrades the build to Node 10.

#### Related Issues / PRs

Part of the work for overleaf/issues#1118

### Review

* Set the Node version in the Dockerfile to 10
* Updated the Dockerfile to build a smaller image
* Added prettier and eslint to `make test`
* Made some manual fixes where decaffeinate got things slightly wrong

There's still some post-decaf ugliness around - I haven't tried much in the way of cleaning up warnings, implicit returns, stuff like that.

#### Potential Impact

The whole of filestore. :bomb: 

#### Manual Testing Performed

- [x] Acceptance tests
- [x] Footle around in the dev env, uploading files and compiling

### Deployment

**High Impact** Do not deploy until January

Remember to test `latexqc`. In staging this is at http://staging-latexqc.ieee.org/


Dockerfile
```diff
@@ -1,23 +1,24 @@
-FROM node:6.9.5 as app
+FROM node:10.17.0-jessie as build
```
briangough: I think Dockerfile comes from the build scripts, so is it going to get overwritten on the next update?

gh2k: Oh, right. I've not really made friends with the build scripts.

Looks like the optimisations I've made won't work in that case.

Actually, it looks like the s3-related things that I put in the other day will get overwritten too.

gh2k:
Updated to use the output of `update_build_scripts`, after adding support for s3 and custom env vars in overleaf/dev-environment#280

I've given up on optimising the `Dockerfile` for size. I can't see a way of circumventing the current build mechanism sufficiently.

das7pad:
The `AWSSDKPersistorManager` would hang similar to #49 due to a `readable` event handler on a stream. I came up with a much smaller patch for that in https://github.com/das7pad/filestore-sharelatex/commit/2957bbceb8adbc5b95cd2a6686117b064d363c33. I could file a PR now or rebase the patch after this PR has landed.

docker-compose.ci.yml
```diff
@@ -22,46 +22,42 @@ services:
...
-
-  fakes3:
+  s3:
     image: adobe/s3mock
     environment:
       - initialBuckets=fake_user_files,fake_template_files,fake_public_files
-    healthcheck:
-      test: ["CMD", "curl", "-f", "http://localhost:9090"]
-
+  healthcheck:
```
das7pad: This looks like a bad indentation?

gh2k:
> The `AWSSDKPersistorManager` would hang 

Thanks. :+1: 

I'm currently starting something of a cleanup that should rationalise the two different talk-to-s3 mechanisms, so I'll make the `readable` changes as part of this work.

docker-compose.ci.yml
gh2k: Thanks. :+1: auto-indentation error in the source jinja files. :/

config/settings.defaults.coffee
```diff
@@ -11,11 +11,11 @@ settings =
 		# Choices are
 		# s3 - Amazon S3
 		# fs - local filesystem
-		if process.env['AWS_KEY']? or process.env['S3_BUCKET_CREDENTIALS']?
```
das7pad: This is backwards incompatible, would you mind adding a fallback on the old environment variable?

buildscript.txt
```diff
@@ -1,10 +1,10 @@
 filestore
---language=coffeescript
---node-version=6.9.5
---acceptance-creds=None
---dependencies=mongo,redis
+--acceptance-creds=
+--public-repo=True
+--language=es
+--node-version=10.17.0
+--dependencies=mongo,redis,s3
```
das7pad: Actually `filestore` does neither depend on mongo nor redis. It was incorrect before, but maybe change it now?

gh2k: Good point :+1:

config/settings.defaults.coffee
gh2k: Yes, that's a wise idea.

PR #62, opened 2019-12-19
Decaf cleanup of SafeExec

gh2k:
### Description

First in a chain of decaf-cleanup PRs. Targeted onto #60 which is the original decaf PR.

This PR cleans up `SafeExec`, and adds support for promises.

#### Related Issues / PRs

Builds on top of #60

PR #63, opened 2019-12-19
Remove unusued S3 bucket-specific credentials and controller

gh2k:
### Description

This is one in a series of decaf-cleanup PRs. This one is based onto #62 and is part of the work started in #60 which is the original decaf PR.

This PR removes `BucketController` and related settings, as they are unused. 

cc @jdleesmiller or @mserranom just in case there's a server pro interaction that I'm not aware of

#### Related Issues / PRs

Original work: #60
Based onto #62


henryoswald:
I think this endpoint was used by the v1 rendered for a publish workflow. I can't remember if this got migrated @jdleesmiller can confirm.

jdleesmiller:
Good catch, Henry. I think this is indeed still used to allow the CLSIs to get publish files from a different bucket. That would be worth testing. Relevant PR: https://github.com/overleaf/filestore/pull/28

gh2k:
Yes, you are correct. It looks like the code that calls into this from v1 is here:
https://github.com/overleaf/write_latex/blob/ccb1b96fc782fe7ac4a6051785efc38f5ea5e03c/main/lib/project_folder.rb#L33-L40

What we're *not* specifying anywhere is the bucket-specific credentials, so I still think these can go. But the controller needs to stay. :+1: 

Will rewrite this PR to drop the setting stuff but keep the controller.

jdleesmiller:
> bucket-specific credentials

If they were anywhere, I think they'd be in the read only filestore for the CLSIs. Are they definitely not used there?

gh2k:
Yeah, they're not used there.

PR #65, opened 2019-12-19
Cleanup and promisify FileConverter

gh2k:
### Description

Part of a series of decaf-cleanup PRs, starting at #60 
This one is targeted onto #64 

This cleans up and promisifies `FileConverter`. As well as promisification, I've also removed some duplicate code present in all of the different conversion methods.

### DIFF
--- a/app/js/FileConverter.js
+++ b/app/js/FileConverter.js
@@ -90,17 +90,6 @@ async function _convert(sourcePath, requestedFormat, command) {
       timeout: FOURTY_SECONDS
     })
   } catch (err) {
-    logger.err(
-      {
-        err,
-        stderr: err.stderr,
-        command,
-        sourcePath,
-        requestedFormat,
-        destPath
-      },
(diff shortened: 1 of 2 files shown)


app/js/FileConverter.js
```diff
@@ -1,133 +1,116 @@
...
+  command.push(destPath)
+  command = Settings.commands.convertCommandPrefix.concat(command)
+
+  try {
+    await safeExec(command, {
+      killSignal: KILL_SIGNAL,
+      timeout: FOURTY_SECONDS
     })
+  } catch (err) {
+    logger.err(
```
ShaneKilkelly: Do we need this log here, if we pass the same info to the error constructor below? Will it be logged fully at the top level anyway? :)

gh2k: No, good call. I think if we're using `OError` we can remove these error logs.

PR #67, opened 2019-12-19
Decaf and promisify ImageOptimiser

gh2k:
### Description

Further cleanup of the decaf started in #60 
This PR cleans up `ImageOptimiser`, and is targeted onto #66 - there are many PRs in this chain as I am trying to spread out the review workload.

Removed the calls directly into `child_process` and instead call directly into `SafeExec`. This also checks that image optimisation is enabled, so removed the redundant check from here.

### DIFF
--- a/app/js/ImageOptimiser.js
+++ b/app/js/ImageOptimiser.js
@@ -1,44 +1,41 @@
-/* eslint-disable
-    no-unused-vars,
-*/
-// TODO: This file was created by bulk-decaffeinate.
-// Fix any style issues and re-enable lint.
-/*
- * decaffeinate suggestions:
- * DS102: Remove unnecessary code created because of implicit returns
- * DS207: Consider shorter variations of null checks
- * Full docs: https://github.com/decaffeinate/decaffeinate/blob/master/docs/suggestions.md
- */
-const { exec } = require('child_process')
 const logger = require('logger-sharelatex')
-const Settings = require('settings-sharelatex')
+const metrics = require('metrics-sharelatex')
+const { callbackify } = require('util')
+const safeExec = require('./SafeExec').promises
 
 module.exports = {
-  compressPng(localPath, callback) {
-    const startTime = new Date()
-    logger.log({ localPath }, 'optimising png path')
-    const args = `optipng ${localPath}`
-    const opts = {
-      timeout: 30 * 1000,
-      killSignal: 'SIGKILL'
+  compressPng: callbackify(compressPng),
+  promises: {
+    compressPng
+  }
+}
+
+async function compressPng(localPath, callback) {
+  const timer = new metrics.Timer('compressPng')
+  logger.log({ localPath }, 'optimising png path')
+
+  const args = ['optipng', localPath]
+  const opts = {
+    timeout: 30 * 1000,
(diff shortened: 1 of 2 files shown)


app/js/ImageOptimiser.js
```diff
@@ -1,44 +1,41 @@
...
+  logger.log({ localPath }, 'optimising png path')
+
+  const args = ['optipng', localPath]
+  const opts = {
+    timeout: 30 * 1000,
+    killSignal: 'SIGKILL'
+  }
+
+  try {
+    await safeExec(args, opts)
```
natestemen: this will return an object with `stdout` and `stderr`, you're just choosing not to collect those values, right?

app/js/ImageOptimiser.js
```diff
@@ -1,44 +1,41 @@
...
+    timer.done()
+    logger.log({ localPath }, 'finished compressing png')
+  } catch (err) {
```
natestemen: any reason for using a `try` `catch` block over doing the `.then` `.catch`?

app/js/ImageOptimiser.js
```diff
@@ -1,44 +1,41 @@
...
- * DS102: Remove unnecessary code created because of implicit returns
- * DS207: Consider shorter variations of null checks
- * Full docs: https://github.com/decaffeinate/decaffeinate/blob/master/docs/suggestions.md
- */
-const { exec } = require('child_process')
 const logger = require('logger-sharelatex')
-const Settings = require('settings-sharelatex')
+const metrics = require('metrics-sharelatex')
+const { callbackify } = require('util')
+const safeExec = require('./SafeExec').promises
```
natestemen: you can rename the function on the fly, right? because you named it differently in https://github.com/overleaf/filestore/pull/62/files#diff-db729691354748e8e0c733b256e0fee2R14

app/js/ImageOptimiser.js
gh2k: That's correct - don't need them here. (Actually, it doesn't look like we use `stdout` anywhere, but it may come in handy.)

app/js/ImageOptimiser.js
gh2k: When using `await` we will receive the 'return' value of the function and it will throw an exception on error... Whereas if you called the method without `await` you would need to `.then` (and `.catch`) instead.

Stylistically, `async`/`await` reads more like Ruby-code than js code.

You can't say `await foo().then()` because you've already resolved the promise with `await`. (Unless `foo` returns a promise that returns a promise.)

app/js/ImageOptimiser.js
gh2k: That's correct, yes. :+1: A function is just an object, so you can assign it to any variable you like. :-)

app/js/ImageOptimiser.js
40thieves: Technically `.then()` will aways return a promise (it's part of the contract of a promise). You can prove this to yourself by running `Promise.resolve('foo').then()` and observing that the return value is a promise.

So you can do `await foo().then()` but there's not really much point you might as well do:

```js
const resolvedValue = await foo()
// do something with resolvedValue
```

PR #68, opened 2019-12-23
Cleanup and promisify acceptance tests. Run tests for all backends.

gh2k:
### Description

This does decaf cleanup for the acceptance tests, and refactors them to run multiple configurations so that they can be run against different backends.

The idea here is that we can modify non-default backends and still test them, enabling us to add more backends in the future.

#### Related Issues / PRs

Initial decaf was #60 
Targeted for #67 

### Review

This required a fairly major refactor of `FilestoreApp` in order to support being torn down and brought up again, including a method to `disrequire` all files from the app directory. - It seems it isn't enough to just disrequire `app.js`, unfortunately.

Renamed the main file from `SendingFileTest` to `FilestoreTests` as this seemed more appropriate. The tests here test all of the functionality, not just sending files. Git doesn't seem to have tracked the rename - I think the files were too different.


test/acceptance/js/FilestoreApp.js
```diff
@@ -1,109 +1,112 @@
...
+        throw err
+      }
+    }
+
+    this.initing = false
+  }
+
+  async waitForInit() {
+    while (this.initing) {
+      await new Promise(resolve => setTimeout(resolve, 1000))
```
emcsween: setTimeout can be promisified via the magic of [util.promisify.custom](https://nodejs.org/dist/latest-v8.x/docs/api/util.html#util_custom_promisified_functions)
```
const sleep = promisify(setTimeout)
await sleep(1000)
```

test/acceptance/js/FilestoreApp.js
```diff
@@ -1,109 +1,112 @@
...
-        ? Settings.filestore.s3.endpoint
-        : undefined)
-    ) {
-      return callback()
+  }
+
+  async stop() {
+    if (this.server) {
+      await new Promise(resolve => {
+        this.server.close(resolve)
```
emcsween: There's no error handling if something goes wrong closing the server. How about:
```
const closeServer = promisify(this.server.close).bind(this.server)
try {
  await closeServer()
} finally {
  delete this.server
}
```

test/acceptance/js/FilestoreApp.js
```diff
@@ -1,109 +1,112 @@
...
-      (err, response) => {
-        console.log(
-          err,
-          response != null ? response.statusCode : undefined,
-          tries
+    let s3Available = false
+
+    while (tries < S3_TRIES && !s3Available) {
+      try {
+        const response = await promisify(request.get)(
```
emcsween: You could use request-promise-native here

test/acceptance/js/FilestoreApp.js
```diff
@@ -1,109 +1,112 @@
...
-        ) {
-          return callback()
+        if ([200, 404].includes(response.statusCode)) {
+          s3Available = true
         }
-
-        if (tries === S3_TRIES) {
-          return callback('timed out waiting for S3')
+      } catch (err) {
+      } finally {
```
emcsween: I would add a comment in the `catch` clause indicating that errors are all swallowed on purpose.

test/acceptance/js/FilestoreApp.js
gh2k: TIL. Thanks :-)

test/acceptance/js/FilestoreApp.js
gh2k: nice :+1:

gh2k:
Good feedback, thanks for the check. :-)

PR #72, opened 2020-01-03
Post-decaf cleanup of app.js

gh2k:
### Description

Cleanup of `app.js`

#### Related Issues / PRs

Started this work in #60 
Targeted onto the branch for #71 

### Review

There is some error-handling middleware that depends on `domain`, which has been deprecated for quite a long time. It looks like this is trying to shutdown gracefully on an uncaught exception. - I extracted this out into `ExceptionHandler.js` to compartmentalise it. I'm not sure I understand exactly what the `domain` stuff is doing, if there's a better way to do this without using a deprecated API, or if we need to do it at all in Node 10. Would appreciate some input. It would be nice to be able to remove the `eslint-disable` for this.

cc @henryoswald, @ShaneKilkelly in case you can shed any light?

There are a couple of other weirdnesses that seem specific to our implementation and I'm not sure how necessary they are. For example, `/shutdown` just sets "not ok", which causes `/status` to return `500` but nothing else, whereas `SIGTERM` actually calls `beginShutdown()`. Is this correct?

Do we need:
```JavaScript
app.get('*', (req, res) => res.sendStatus(404))
```
or is this implicit?


app.js
```diff
@@ -180,68 +131,50 @@ app.get('/heapdump', (req, res, next) =>
...
 
 app.get('/status', function(req, res) {
-  if (appIsOk) {
-    return res.send('filestore sharelatex up')
+  if (exceptionHandler.appIsOk()) {
+    res.send('filestore sharelatex up')
   } else {
     logger.log('app is not ok - shutting down')
-    return res.send('server is being shut down', 500)
+    res.send('server is being shut down').status(500)
```
das7pad:
```suggestion
    res.status(500).send('server is being shut down')
```

`.send` is adding a default status code of 200 and calling `.end` with the payload. In order to set a different status code, we need to call `.status` first.

app/js/ExceptionHandler.js
```diff
@@ -0,0 +1,97 @@
...
+  appIsOk() {
+    return this._appIsOk
+  }
+
+  setNotOk() {
+    this._appIsOk = false
+  }
+
+  middleware(req, res, next) {
+    const rescueLogger = require('logger-sharelatex')
```
das7pad: Why can't we use the regular `logger` here?

app/js/ExceptionHandler.js
```diff
@@ -0,0 +1,97 @@
...
+    const requestDomain = domain.create()
+    requestDomain.add(req)
+    requestDomain.add(res)
+    requestDomain.on('error', err => {
+      try {
+        // request a shutdown to prevent memory leaks
+        this.beginShutdown()
+        if (!res.headerSent) {
+          res.send('uncaught exception').status(500)
```
das7pad:
```suggestion
          res.status(500).send('uncaught exception')
```
see comment above.

app/js/ExceptionHandler.js
gh2k: Good question. Primarily I'm copying the pattern that was already here, in that the logger was previously also re-required in the error handler. This could well have been bad code in the first place, rather than on purpose, back in the original CoffeeScript.

My assumption was that as this is effectively a global exception handler, we might have stuffed up the logger object and would need a new one, but I'm not sure whether re-requiring it here would be that helpful, considering what happens with caching and whatnot.

@briangough originally made the change to include the new logger in https://github.com/overleaf/filestore/commit/6842c7dd0a42acdcafdb70bb6be40515192db2ff so this looks like it was done on purpose, but 2014 was a long time ago and perhaps the reason for this is no longer relevant. :wink:

gh2k:
Thanks for the comments, @das7pad - good catch

app/js/ExceptionHandler.js
das7pad: It was even part of the initial commit https://github.com/overleaf/filestore/commit/5881a3ea2c913d5e004199916d4baa99f29f2c1e#diff-66ac8d4b3a5e0f5ee69c82648d9ade91R39.

There is just one `logger-sharelatex` object, which will be pulled in via the second `require`. Calling `.initialize` resets parts of it - creating a new `Bunyan` logger instance and streams. But we are not calling `.initialize` here.

So I would opt for removal here.

gh2k: Yeah, I think you're right. I'd missed that it was there at the beginning. :+1:

henryoswald:
Yep domain was at one point the future way node would handle exceptions, then they realised it was crap so we stopped adding it. 

`app.get('*', (req, res) => res.sendStatus(404))` it is probably implicit but worth testing it doesn't return a different status code.

gh2k:
Nice. Removed all of the unnecessary stuff. I verified that it does indeed return 404 by default.

PR #73, opened 2020-01-04
Cleanup and refactor S3PersistorManager to use aws-sdk only

gh2k:
### Description

Refactor (or 'rewrite' might be a more sensible term) `S3PersistorManager` to use only `aws-sdk` (taking some inspiration from `AWSSDKPersistorManager`,) and promisifying.

#### Related Issues / PRs

Started this work in #60 and targeted onto #72 

### Review

From the git history, it looks like `AWSSDKPersistorManager` was implemented but never made it into production use. I can't find an associated PR so I'm not 100% sure of the history, but it looks like this is the case.

Removed the `AWSSDKPersistorManager` and configured filestore to use the rewritten code when `aws-sdk` is configured.

I realise that this is a tough one to review because so much has changed. This is the main reason why I wanted to sort out the acceptance tests before doing this. - It is currently hard to test as the dev env doesn't support S3 beyond the unit tests. It might be worth looking into this, but at the same time, we want to migrate this to use GCS which may be a better thing to spend effort setting up.

I think at this point we're probably in the realm of "test it very thoroughly in staging" as the main way of verifying that it works.


app/js/FileController.js
```diff
@@ -61,7 +61,8 @@ function getFile(req, res) {
     }
 
     logger.log({ key, bucket, format, style }, 'sending file to response')
-    pipeline(fileStream, res)
+
+    pipeline(fileStream, res, next)
```
briangough: maybe add the commit message as a comment here, it wasn't immediately obvious that `next` was being used as the callback.

PR #74, opened 2020-01-07
Restore bucket-specific settings and clean up related endpoint

gh2k:
### Description

This restores the bucket-specific settings to `S3PerisstorManager`. Originally I removed these in #63, but it turns out this functionality was needed so I've cleaned it up instead.

Also tidies up the bucket-specific file access endpoint, to use the same mechanism as the other file-getting endpoints, by applying a different key builder middleware and using `FileController`.

This *should* be the last decaf/cleanup PR. :sweat_smile: 

#### Related Issues / PRs

Started the work in #60 
~Targeted onto #73~ - Merged the other PRs into the integration branch, so now this is targeted onto #60

### Review

The only thing that `BucketController` did differently was to load the bucket-specific settings. Added these into `S3PersistorManager` instead.

Added an acceptance test for the endpoint, which puts a file to access in a random bucket first. This requires path-style access in the s3 client to access the bucket.


app/js/S3PersistorManager.js
```diff
@@ -240,9 +249,49 @@ function _wrapError(error, message, params, ErrorType) {
   }
 }
 
-function _defaultOptions() {
-  const options = {
-    credentials: {
+const _clients = {}
+
+function _client(bucket) {
```
henryoswald: pedantic comment you can ignore, it isn't very clear what `_client()` does, would `_getClient` be a touch more descriptive?

gh2k: I've always had a soft preference for omitting `get` where possible, e.g. `foo.client()`, `foo.setClient(bar)`

Although in this case, I think it might be more descriptive to call it `_clientForBucket(bucket)`

It's definitely not the thing I reserve my strongest opinions for, though.

app/js/S3PersistorManager.js
```diff
@@ -240,9 +249,49 @@ function _wrapError(error, message, params, ErrorType) {
   }
 }
 
-function _defaultOptions() {
-  const options = {
-    credentials: {
+const _clients = {}
```
jdleesmiller: Could this use a `Map`? I think it is best to move away from using objects as maps where possible, now that we have maps.

app/js/S3PersistorManager.js
```diff
@@ -240,9 +249,49 @@ function _wrapError(error, message, params, ErrorType) {
...
+    settings.filestore.s3.s3BucketCreds[bucket]
+  ) {
+    _clients[bucket] = new S3(
+      _clientOptions(settings.filestore.s3.s3BucketCreds[bucket])
+    )
+    return _clients[bucket]
+  }
+
+  // no specific credentials for the bucket
+  if (_clients.default) {
```
jdleesmiller: This is very slightly dangerous, because a bucket named `default` would get this client. How would it look if the default client was in its own variable?

app/js/S3PersistorManager.js
jdleesmiller: Well, since we're talking about it... I usually try to make all function names verbs, so something like `getClientForBucket`.

It may also be worth noting that it's now relatively easy to define custom getters/setters in JS, e.g. to make `foo.bar = foo.bat` work even if `bar` and `bat` have some logic in them, which can cut down on a lot of the `get` and `set` noise for really simple getters and setters, which is I think one of the motivations for avoiding `get`.

app/js/S3PersistorManager.js
gh2k: Yes, this sounds like a good habit to cultivate.

app/js/S3PersistorManager.js
gh2k: I suspect that the person who owns the bucket named `default` won't be using this code, but you're right nonetheless.

gh2k:
Addressed comments and merging into integration branch

PR #75, opened 2020-01-07
Simplify logging to log only once per http request

gh2k:
### Description

An attempt to reduce log volume to (mostly) one log per request, as described in the logging proposal. Specifically, this does the following:

* Removes access to the logger from (almost) everywhere
* Adds a logging middleware to express, and a logging error handler
* Call either `next()` or `next(err)` from all non-health-check controller methods

This results in one happy-path log per request, *or* one error log.

#### Related Issues / PRs

This is currently hanging off my daisy-chain of branches, so currently points at #74 
The logging proposal is in overleaf/write_latex_ops#531

### Review

This is something of a draft/out-of-hours PR so I'm happy to accept bikeshedding or outright rejection here. :-)

The logging middleware allows us to write some context to the `res` object, which it will read before printing anything. This enables customisation of the log message.

If anything goes wrong, we wind up with an `OError` which we pass to `next(err)`, which will log the error and return `500` with the message from the error. This allows us to populate it with plenty of context, and removes the need for adding multiple error logs when things go awry.

Because of this, very little needs to `require('logger-sharelatex')`. One place where I *have* done this is in the `ImageOptimiser` as it otherwise swallows errors when the optimisation process ties out. This seems sensible behaviour as we already have a non-optimised png to serve, but it might be a useful thing to know about. - Another place is when initialising the `PersistorManager` as it may be helpful to have a log line saying what backend it's trying to boot.

This is one approach to the logging proposal and I'm sure there are plenty of others.

### DIFF
--- a/app/js/RequestLogger.js
+++ b/app/js/RequestLogger.js
@@ -1,0 +1,32 @@
+const logger = require('logger-sharelatex')
+
+module.exports = {
+  logRequest,
+  logError
+}
+
+function logRequest(req, res) {
+  // response has already been sent, but we log what happened here
+  logger.log(
+    {
+      info: res.logInfo,
+      url: req.originalUrl,
+      params: req.params
+    },
+    res.logMsg || 'HTTP request'
+  )
+}
+
+function logError(err, req, res, next) {
+  logger.err(
+    {
+      err,
+      info: res.logInfo,
+      url: req.originalUrl,
+      params: req.params,
+      msg: res.logMsg
+    },
+    err.message
+  )
+  next(err) // use the standard error handler to send the response
+}
--- a/app.js
+++ b/app.js
@@ -12,6 +12,8 @@ const fileController = require('./app/js/FileController')
 const keyBuilder = require('./app/js/KeyBuilder')
 const healthCheckController = require('./app/js/HealthCheckController')
 
+const RequestLogger = require('./app/js/RequestLogger')
+
 const app = express()
 
 if (settings.sentry && settings.sentry.dsn) {
@@ -27,6 +29,7 @@ if (Metrics.event_loop) {
 app.use(Metrics.http.monitor(logger))
 app.use(function(req, res, next) {
   Metrics.inc('http-request')
+  res.logInfo = {}
   next()
 })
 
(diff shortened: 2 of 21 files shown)


gh2k:
force-pushed to resolve conflicts with the base branch

gh2k:
Currently this logs twice per request, as `Metrics.http.monitor` actually logs a line as well a recording timing metrics.

I think a solution here would be to remove this monitor and add the http timing stuff to `RequestLooger`, although it does seem slightly awkward that the one metrics method does both things.

40thieves:
In general I like the approach. 

My main concern is mostly that we're hanging random properties off `req`. If you typo the property or accidentally assign over `req.logInfo` then data could be lost. I think this could be improved if manipulation of these properties was via a generic set of methods (something like `req.appendLogInfo({ ... })`)

This feels like something that could be solved with a generic middleware, and possibly someone has already written one (although I wasn't able to find anything with some cursory searches).

I also note that in the proposal, Technical Proposal 2 mentions using `req.log.LOG_LEVEL()` on `logging-bunyan` would bundle log messages together in Stackdriver. However, I'm not sure if a) this does exactly what we want and b) how this would work with CE/SP.

app/js/RequestLogger.js
```diff
@@ -0,0 +1,32 @@
+const logger = require('logger-sharelatex')
+
+module.exports = {
+  logRequest,
+  logError
+}
+
+function logRequest(req, res) {
+  // response has already been sent, but we log what happened here
+  logger.log(
```
emcsween: the log level could depend on the status code here.

gh2k:
Thanks for the feedback, everyone. :bowing_man: 

It looks like the general consensus is to add a helper method, which I approve of. :+1: Indeed, using `req` vs `res` was a typo on my part.

> removing the trace logs

Yeah, my feeling is that if everything went to plan, we should be able to infer what will happen from the request parameters. So, we probably only need more information on what happened if there is an error.

> `@google-cloud/logging-bunyan`

I do like that approach as a way of keeping them. If `filestore` were wider in scope I'd look into this here, but I think the approach taken in this PR is probably sufficient for what we need, and cleaner than what we have at the moment.

> This PR removes some code that sanitizes credentials in logs. 

That was on purpose - we're not passing any credentials via `opts` any more as this logic was localised just to the S3 persistor as part of the cleanup. But I agree that we should still be aware of when credentials might be around.

I'll make the changes suggested here and move the PR out of draft.

PR #77, opened 2020-01-10
Simplify logging to only log once per request

gh2k:
**Re-opening of #75**

Accidentally closed that PR by cleaning up some stale branches, and now I can't reopen it again.

### Description

An attempt to reduce log volume to (mostly) one log per request, as described in the logging proposal. Specifically, this does the following:

* Removes access to the logger from (almost) everywhere
* Adds a logging middleware to express, and a logging error handler
* Call either `next()` or `next(err)` from all non-health-check controller methods

This results in one happy-path log per request, *or* one error log.

#### Related Issues / PRs

Previously opened as #75
The logging proposal is in overleaf/write_latex_ops#531

### Review

In addition to the comments made in the previous PR, I took the logic from `metrics-sharelatex`'s HTTP monitor, which records timing info for the http methods as well as acting as a request logger itself. This avoids getting two messages per request.

Additionally, this logic (which overrides `res.end`) means that you don't have to call `next` in the controller methods except in the case of an error, which is an improvement. (`end` is always called by the framework even if you don't call it explicitly)

Force-pushed the branch to resolve a conflict with master, but the only changes from the previous PR are in the second commit here.

### DIFF
--- a/app.js
+++ b/app.js
@@ -12,6 +12,8 @@ const fileController = require('./app/js/FileController')
 const keyBuilder = require('./app/js/KeyBuilder')
 const healthCheckController = require('./app/js/HealthCheckController')
 
+const RequestLogger = require('./app/js/RequestLogger')
+
 const app = express()
 
 if (settings.sentry && settings.sentry.dsn) {
@@ -27,6 +29,7 @@ if (Metrics.event_loop) {
 app.use(Metrics.http.monitor(logger))
 app.use(function(req, res, next) {
   Metrics.inc('http-request')
+  res.logInfo = {}
   next()
 })
 
@@ -136,6 +139,9 @@ app.get('/status', function(req, res) {
 })
 
 app.get('/health_check', healthCheckController.check)
+
+app.use(RequestLogger.logRequest)
+app.use(RequestLogger.logError)
 
 const port = settings.internal.filestore.port || 3009
 const host = '0.0.0.0'
--- a/app.js
+++ b/app.js
@@ -26,7 +26,9 @@ if (Metrics.event_loop) {
   Metrics.event_loop.monitor(logger)
 }
 
-app.use(Metrics.http.monitor(logger))
+app.use(RequestLogger.middleware)
+app.use(RequestLogger.errorHandler)
+
 app.use(function(req, res, next) {
   Metrics.inc('http-request')
   res.logInfo = {}
@@ -139,9 +141,6 @@ app.get('/status', function(req, res) {
 })
 
 app.get('/health_check', healthCheckController.check)
-
-app.use(RequestLogger.logRequest)
-app.use(RequestLogger.logError)
 
 const port = settings.internal.filestore.port || 3009
 const host = '0.0.0.0'
--- a/app.js
+++ b/app.js
@@ -16,8 +16,7 @@ const RequestLogger = require('./app/js/RequestLogger')
 
 const app = express()
 
-const requestLogger = new RequestLogger()
-requestLogger.attach(app)
+RequestLogger.attach(app)
 
 if (settings.sentry && settings.sentry.dsn) {
   logger.initializeErrorReporting(settings.sentry.dsn)
--- a/app.js
+++ b/app.js
@@ -16,6 +16,9 @@ const RequestLogger = require('./app/js/RequestLogger')
 
 const app = express()
 
+const requestLogger = new RequestLogger()
+requestLogger.attach(app)
+
 if (settings.sentry && settings.sentry.dsn) {
   logger.initializeErrorReporting(settings.sentry.dsn)
 }
@@ -25,9 +28,6 @@ Metrics.memory.monitor(logger)
 if (Metrics.event_loop) {
   Metrics.event_loop.monitor(logger)
 }
-
-app.use(RequestLogger.middleware)
-app.use(RequestLogger.errorHandler)
 
 app.use(function(req, res, next) {
   Metrics.inc('http-request')
--- a/app.js
+++ b/app.js
@@ -30,7 +30,6 @@ if (Metrics.event_loop) {
 
 app.use(function(req, res, next) {
   Metrics.inc('http-request')
-  res.logInfo = {}
   next()
 })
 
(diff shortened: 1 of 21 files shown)


app/js/RequestLogger.js
```diff
@@ -0,0 +1,83 @@
...
+}
+
+function middleware(req, res, next) {
+  const startTime = new Date()
+
+  // methods to allow the setting of additional information to be logged for the request
+  req._logInfo = {}
+  req._logMessage = 'http request'
+  req.addLogField = function(field, value) {
+    req._logInfo[field] = value
```
40thieves: This is a mega-nitpick, but I think I would probably just use `req.addLogFields({ foo })` (i.e. object shorthand for `{ foo: foo }` over the separate key/value arguments, therefore I'd argue for removing this. YMMV though.

app.js
```diff
@@ -24,9 +26,12 @@ if (Metrics.event_loop) {
   Metrics.event_loop.monitor(logger)
 }
 
-app.use(Metrics.http.monitor(logger))
+app.use(RequestLogger.middleware)
+app.use(RequestLogger.errorHandler)
+
 app.use(function(req, res, next) {
   Metrics.inc('http-request')
+  res.logInfo = {}
```
40thieves: Think this should be removed now.

app/js/RequestLogger.js
```diff
@@ -0,0 +1,83 @@
...
+  req.addLogFields = function(fields) {
+    Object.assign(req._logInfo, fields)
+  }
+  req.setLogMessage = function(message) {
+    req._logMessage = message
+  }
+
+  // override the 'end' method to log and record metrics
+  const end = res.end
+  res.end = function() {
```
40thieves: I am a tiny bit uneasy with monkey patching `res.end`, but equally I can't think of a specific argument against. I wonder if it should be done in a separate middleware, but that might be overkill.

I'd be interested in hearing other people's opinions.

app/js/RequestLogger.js
gh2k: Yes, good point. Two ways of doing the same thing. Happy to remove it.

app/js/RequestLogger.js
gh2k: Yeah - it's a bit on the ugly side but this is what `metrics-sharelatex` does, so we've been using it for some time. There are two reasons to do it this way:
* It allows us to keep the timing stuff that `metrics-sharelatex` does
* It ensures the logging happens on every request, without having to make sure we call `next()` in every controller method.

AFAIK `express` doesn't otherwise provide any kind of 'the request has now been processed' event.

app/js/RequestLogger.js
```diff
@@ -1,32 +1,83 @@
...
+  req._logMessage = 'http request'
+  req.addLogField = function(field, value) {
+    req._logInfo[field] = value
+  }
+  req.addLogFields = function(fields) {
+    Object.assign(req._logInfo, fields)
+  }
+  req.setLogMessage = function(message) {
+    req._logMessage = message
+  }
```
emcsween: A little code organization suggestion:

`_logInfo`, `_logMessage`, `addLogFields` and `setLogMessage` implement together the logging feature. How about bringing them together in an object?
```
class RequestLogger {
  constructor() {
    this._logInfo = {}
    this._logMessage = 'http request'\
  }
  addFields(...) { ... }
  setMessage(..) { ... }
  emitLog(...) {
    logger.log(this._logMessage, ...)
  }
}

function middleware(req, res, next) {
  ...
  req.logger = new RequestLogger()
  ...
  res.end = function() {
    ...
    req.logger.emitLog()
    ...
  }
}
```

gh2k:
> The metrics middleware and the logging middleware are independent. They could be implemented as separate middlewares (which would nicely fit the sharelatex-logger and sharelatex-metrics separation)

We end up with a double-override of the `end` method, but I guess if it's good enough to do once, it's good enough to do twice. ;-)

> When this PR is approved, the middlewares could be moved back to the sharelatex-logger and sharelatex-metrics repos. The logger is backwards compatible. If the setLogMessage and addLogFields methods are not called, requests will be logged the same way as they currently are.

Are you suggesting that we patch `logger.log` here to talk to the `RequestLogger`? That seems a much wider scoped change. I would currently consider the existing log methods as independent of this mechanism.

The problem with merging this into Metrics as it currently stands is that the Metrics http monitor currently does request logging, and splitting the functionality and then merging would break that. That would break other things that are using this.

gh2k:
Removed @jdleesmiller from reviewers because he's very busy, but cc here because he wrote the logging proposal. :-)

emcsween:
> We end up with a double-override of the `end` method, but I guess if it's good enough to do once, it's good enough to do twice. ;-)

Right. It might even be overridden more than that by other middlewares. On the other hand, I now realize that the response time is also written to the logs, so this is a bad idea after all...

> Are you suggesting that we patch `logger.log` here to talk to the `RequestLogger`?

No, I'm suggesting just making the middleware available to other apps. If I understand correctly, if it's just plugged in without modifications to the app, it will behave the same as before, logging "http request" with some request info for each request? Having it in sharelatex-metrics would allow us to start using the new features incrementally in other apps.

40thieves:
I think I agree with Eric, this feels more generic than just filestore. However perhaps it's worth trying out here first before we start generalising it (further)?

app/js/RequestLogger.js
```diff
@@ -1,83 +1,90 @@
...
   }
 
-  // override the 'end' method to log and record metrics
-  const end = res.end
-  res.end = function() {
-    // apply the standard request 'end' method before logging and metrics
-    end.apply(this, arguments)
+  middleware(req, res, next) {
+    const startTime = new Date()
+    req.requestLogger = this
```
emcsween: Neat idea to avoid creating a separate object for each request, but this won't work with concurrent requests, right? They'll all race to modify the same _logMessage and _logInfo?

gh2k: You are correct, this code is nonsense. I'll change it. :+1:

gh2k:
I think this is good to go at this point. :crossed_fingers:

gh2k:
>  if we know the request path and method, that usually contains the same information

I thought about this and figured that the messages might be useful when reading manually through events. It *might* help to see
- creating wombat
- creating potato
- feeding potato to wombat
to save having to look up what each of those routes do, when debugging something.

PR #78, opened 2020-01-16
Add Migration Persistor, to send missing file requests to a fallback persistor

gh2k:
### Description

Creates a new `MigrationPersistor`, which:
- Is initialised with 'primary' and 'fallback' persistors
- Sends all writes to the primary
- Reads from the fallback when the primary is not available
- *Optionally* copies files from the fallback to the primary, when they are read
    - This takes place parallel to sending the file to the client, so as not to make this process any slower
- Sends delete requests to both

#### Related Issues / PRs

Part of the work for overleaf/google-ops#731

### Review

There's a noisy commit first that simply renames `FSPersistorManager` and `S3PersistorManager` to just `FSPersistor` and `S3Persistor`. The `PersistorManager` is the thing that manages these things, so I thought that this would be a better naming convention. I think it would be wise to review these commits separately.

Have beefed up the acceptance tests quite heavily.

Added a metric to log file copies that occur as a result of a miss on the primary. This should help us track how well the migration is going.

#### Potential Impact

Shouldn't have any appreciable change unless we change the settings to enable this mechanism, so it should be fairly safe to go out.


gh2k:
It's possible to use `minipass` to tee the source stream to the destination and the client at the same time, which would indeed save bandwidth. `streams.PassThrough` doesn't really work because of the way it handles clients consuming the data at different speeds. (badly) 

I haven't checked what happens if the client aborts the transfer with 'minipass' though. - I don't know if it will also abort the write to the primary, or continue with stream `a` if `b` goes away. (Reading the docs it looks like the latter, but we should test this if we can.)

Alternatively, we could fetch the file to disk and then pipe it to the client but I'm reluctant to do this as it will slow down the transfer, as it will have to download the entire file before sending it to the client.

WRT file verification, there are some options.

The simplest thing to do is check the file size. This protects us against the 'we wrote a zero-byte' file problem, or 'the upload got aborted', which I think are the two most likely cases.

For 'the file is the right size but the data is bad', we can:

- Fetch the download, hash it, upload it, download it again, hash it
    - This does not protect against the data being wrong on the first download, so if we fetch some bad data and upload it, we still have bad data
- Use the hash metadata
    - S3's ETag header *should* be an md5 hash. They say that this will be the case for standard objects sent via `putObject`, but not for multipart uploads. AFAIK we have never done multipart uploads, so I think this will be fine
    - Google provide explicit `md5` metadata

I think the latter is better, but as the etag isn't *guaranteed* to be the md5 hash (although I think it will be) it's slightly awkward. - We could keep tabs on this with a metric though. If we're going to clean up the destination file, not fail and log a warning, it should be fairly safe to do this and keep track of when things go wrong.

WDYT?

gh2k:
Discussed in call, I'll try this:

- Pipe stream from S3 to GCS (and client, if necessary)
- Compare MD5 based on S3.etag and GCS.md5 metadata
- Delete the file from GCS if they don't match, log a warning and increment a metric

It's not straightforward to write the `FileHashManager` hash on upload, as this requires knowing the file size in advance of the write. (See https://github.com/overleaf/web-internal/blob/be46affcc95b7e4c79f86503da933c6fcc277d4f/app/src/Features/FileStore/FileHashManager.js#L51) - This isn't compatible with the `sendStream` method.

briangough:
Using the hash metadata is good, if we pass the header to google I think we can rely on them to check it  [reference-headers#xgooghash](https://cloud.google.com/storage/docs/xml-api/reference-headers#xgooghash) (so we don't need to delete the file ourselves if the hash doesn't match).  

Can we get the content-length on incoming streams? I think that might be possible when we upload a file, as it is uploaded to disk on web. If not perhaps [File.js#L18-L20](https://github.com/overleaf/web-internal/blob/c9c25e09f29804a7f087cba0007bb0e285c60650/app/src/models/File.js#L18-L20) needs to be extended to record the file size and/or the md5.

If we extend filestore to do caching of downloaded files in persistent storage that might give us the opportunity to check the hash more easily.  Otherwise we should  try to check it on the stream at least and log any error.

test/acceptance/js/FilestoreTests.js
```diff
@@ -320,14 +397,302 @@ describe('Filestore', function() {
...
+                constantFileContent
+              )
+            })
+
+            it('should fetch the file', async function() {
+              const res = await rp.get(fileUrl)
+              expect(res.body).to.equal(constantFileContent)
+            })
+
+            it('should not copy the file to the primary', async function() {
```
das7pad: Can you wrap this case in a 'when copyOnMiss is disabled' suite?

Maybe with an explcit beforeEach handler `Settings.filestore.fallback.copyOnMiss = false`

It is not obvious from the nearby context. And we might want to change this default at some point.

test/acceptance/js/FilestoreTests.js
```diff
@@ -320,14 +397,302 @@ describe('Filestore', function() {
...
+                  bucket,
+                  fileKey,
+                  constantFileContent
+                )
+              })
+            })
+
+            describe('when copying a file', function() {
+              let newFileId, newFileUrl, newFileKey
+              const newProjectID = 'acceptance_tests_copied_project'
```
das7pad: Let's use something random here. 'acceptance_tests_copied_project' is used in line 245 already.

test/acceptance/js/FilestoreTests.js
```diff
@@ -320,14 +397,302 @@ describe('Filestore', function() {
...
+                fileKey,
+                constantFileContent
+              )
+            })
+
+            it('should not store the file on the fallback', async function() {
+              await expectPersistorNotToHaveFile(
+                app.persistor.fallbackPersistor,
+                fallbackBucket,
+                `acceptance_tests/${directoryName}/${fileId}`
```
das7pad:
```suggestion
                `${projectId}/${directoryName}/${fileId}`
```

gh2k:
Thanks for the review @das7pad - good points :-)

> if we pass the header to google I think we can rely on them to check it reference-headers#xgooghash (so we don't need to delete the file ourselves if the hash doesn't match).

:+1: 

> Can we get the content-length on incoming streams? I think that might be possible when we upload a file, as it is uploaded to disk on web. If not perhaps File.js#L18-L20 needs to be extended to record the file size and/or the md5.

I'm not very keen on the idea of changing the external API for `filestore`, which is what we will have to do to support the extra parameters. It increases the scope of the change quite a bit and introduces more scope for error on what is already quite a big change.

I think putting the md5 hash in `File` is a good idea though. That will let us compare what Mongo thinks should be there with what Google thinks is there.

> we should try to check it on the stream

Perhaps we can add md5 validation in a later patch for incoming streams. Are you suggesting that we check the md5 on all downloads here? I think it's doable - we would just need to add some stream events - but it would increase the CPU load on filestore-readonly particularly, which is used quite heavily by the clsis.

jdleesmiller:
One thought on copy-on-miss: do we need that feature? If it is causing some complexity with the streaming, I think we could live without it, since we'll also presumably have a migration script that can copy the misses, and we can presumably set that off starting with more recent projects and working backward, which should have a similar effect. WDYT?

gh2k:
> One thought on copy-on-miss: do we need that feature? If it is causing some complexity with the streaming, I think we could live without it, since we'll also presumably have a migration script that can copy the misses, and we can presumably set that off starting with more recent projects and working backward, which should have a similar effect. WDYT?

Well, it's almost implemented now but it does add complexity. The thinking behind copy on miss is that it will save traffic for projects that are seeing use, and thus prevent us from having to fetch things multiple times when they are used. It should pull most of the data across when the project is opened and the first compile request goes off.

Working backwards through projects comes with its own set of complexities because it needs its own resume logic (which I imagine we will have to do many times) and it prevents us from using something like the [storage transfer service](https://cloud.google.com/storage-transfer/docs/overview) which I think would be preferable to using a script. I think this would minimise the bandwidth going through our VPC, and wouldn't need babysitting.

gh2k:
I've made the changes requested and added the md5-validation. Also uses `minipass` to read only once from the fallback and pipe to both the client and the primary, when copying.

Sorry that this is such a big review. I'm not sure how to turn it into more bitesize chunks.

gh2k:
I've un-requested reviews for now - I'm refactoring some of this as part of the work for the GCS persistor to reduce code reuse. Feedback still welcome if you've started looking though.

gh2k:
Looking at it, the refactoring is all being done in S3Persistor, so re-opening this as the refactoring work isn't relevant.

Sorry about that.

test/unit/js/MigrationPersistorTests.js
```diff
@@ -13,43 +13,61 @@ const Errors = require('../../../app/js/Errors')
 // deleteFile: _wrapMethodOnBothPersistors
 // copyFile: copyFileWithFallback
 
-describe('MigrationPersistorTests', function() {
+describe.only('MigrationPersistorTests', function() {
```
das7pad: Let's run all the test suites again :wink:

app/js/FSPersistor.js
```diff
@@ -80,6 +92,38 @@ async function getFileSize(location, filename) {
...
+      const hash = crypto.createHash('md5')
+      hash.setEncoding('hex')
+      readStream.on('end', () => {
+        hash.end()
+        resolve(hash.read())
+      })
+      readStream.on('error', err => {
+        throw err
+      })
+      readStream.pipe(hash)
```
das7pad: In order to catch the `err` of the `readStream.on('error'...)` handler in the `catch` block below, we need to use `Stream.pipeline` here, no?

test/unit/js/MigrationPersistorTests.js
gh2k: Thanks. `eslint` was supposed to catch this but `.eslintrc` is incorrect. WIll fix this here.

app/js/MigrationPersistor.js
```diff
@@ -68,46 +69,158 @@ module.exports = function(primary, fallback) {
...
+
+      await primary.promises.sendStream(destBucket, destKey, stream, sourceMd5)
+    } catch (err) {
+      let error = err
+      metrics.inc('fallback.copy.failure')
+
+      try {
+        await primary.promises.deleteFile(destBucket, destKey)
+      } catch (err) {
+        error = new WriteError({
```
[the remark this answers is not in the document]
gh2k: I'm trying to avoid unnecessary log lines if possible. The `withCause` will wrap the error, and the error message made it clear what went wrong here so I think this is OK. It gives us a little extra context without having to resort to another `logger.warn` at this point.

app/js/FSPersistor.js
gh2k: Yeah, the format of this method is a bit weird. There's not really a need to try/catch at all - can just use reject on error if pipeline fails.

app/js/MigrationPersistor.js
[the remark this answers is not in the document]
das7pad: To be more specific, the error catched from line 115 is lost.

The error message seen outside `_copyFromFallbackStreamAndVerify` may be simplified as
```
unable to copy file to destination persistor
withCause: unable to clean up destination copy artifact
withCause: some PRIMARY.deleteFile error
```
But this is more important:

```
unable to copy file to destination persistor
withCause: some PRIMARY.sendStream error
withCause: ...
```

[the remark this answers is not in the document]
gh2k: Oh, yes I see what you mean.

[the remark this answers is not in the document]
gh2k: Swizzled things around to make the cleanup error a property of the main error. :+1:

test/acceptance/js/FilestoreTests.js
```diff
@@ -320,14 +397,307 @@ describe('Filestore', function() {
...
+
+              it('should not create a new file in the old bucket', async function() {
+                await expectPersistorNotToHaveFile(
+                  app.persistor.fallbackPersistor,
+                  fallbackBucket,
+                  newFileKey
+                )
+              })
+
+              it('should not copy the old file to the new bucket', async function() {
```
das7pad: What is the rational for that?

A followup copy call in the primary persistor would save us another download from the fallback + upload to the primary.

app/js/MigrationPersistor.js
```diff
@@ -0,0 +1,221 @@
...
+    ).catch(
+      // the error handler in this method will log a metric and a warning, so
+      // we don't need to do anything extra here, but catching it will prevent
+      // unhandled promise rejection warnings
+      () => {}
+    )
+
+    // start piping the source stream into the tee after everything is set up,
+    // otherwise one stream may consume bytes that don't arrive at the other
+    sourceStream.pipe(tee)
```
briangough: Is there a reason we can't do this with normal streams?  I am not sure why Minipass is needed.  If the code is executed as
```
sourceStream.pipe(clientStream)
sourceStream.pipe(copyStream)
```
there shouldn't be a problem with data being lost because the sourceStream will not start flowing until the next tick (based on the presence of the `process.nextTick` in the `resume` method [_stream_readable.js#L949](https://github.com/nodejs/node/blob/43fb6ffef7a14b47d1b0ba8134102795f5017a59/lib/_stream_readable.js#L949) which is called by the `pipe` method at [_stream_readable.js#L780](https://github.com/nodejs/node/blob/43fb6ffef7a14b47d1b0ba8134102795f5017a59/lib/_stream_readable.js#L780) ->  [_stream_readable.js#L940](https://github.com/nodejs/node/blob/43fb6ffef7a14b47d1b0ba8134102795f5017a59/lib/_stream_readable.js#L940)  )

gh2k: Right, so I read something somewhere when first doing some googling that nodejs streams don't handle multiple pipe destinations well if one is a lot slower than the other and can result in a buffer getting overwritten.

I cannot find this information again, so either it was very old or I dreamt it, but I have been labouring under the misapprehension that you should use a passthrough stream if you want to 'tee' stuff from a to b.

In fact, the top answer for nodejs tee streams explains that this is indeed a comon pattern.
omg

gh2k:
Indeed. I shall revise. :man_facepalming:

test/acceptance/js/FilestoreTests.js
gh2k: Good question. If `copyOnMiss` is enabled it's probably a good idea to copy it to both locations. :+1:

gh2k:
OK, so I have:

* Removed the dependency on minipass
* Used `Stream.PassThrough` where still necessary
    * Refactored to make it cleaner and clearer why this is necessary
    * Removed the unnecessary passthrough logic from `S3Persistor`
* Changed the behaviour of copy when `copyOnMiss` is true, to copy the file to both the old and new locations on the primary

The reason why we still need to use pass-through streams is that, depending on what wants to consume data from the stream, we can end up letting one thing read data from the stream before the `pipe`/`pipeline` has been set up for the other listeners. This is not always the case, depending on whether the client attaches to the stream straight away and if the server wants to calculate the md5 hash of the stream, but logically it makes more sense to split the stream into two at the start. - It's safer and the caller doesn't need to know about any special circumstances.

This could be achieved with `Stream.PassThrough` however, so getting rid of `minipass` is still worth doing to reduce the number of dependencies that we have.

Removed the unnecessary `tee` streams that were just overengineering it.

app/js/FSPersistor.js
```diff
@@ -80,6 +92,31 @@ async function getFileSize(location, filename) {
...
+    const hash = crypto.createHash('md5')
+    hash.setEncoding('hex')
+    readStream.on('end', () => {
+      hash.end()
+      resolve(hash.read())
+    })
+    pipeline(readStream, hash).catch(reject)
+  })
+}
+
```
briangough: This could be simplified a bit by returning the promise from the promisified `pipeline` instead of creating a new promise.
```
async function _getFileMd5HashForPath(fullPath) {
  const readStream = fs.createReadStream(fullPath)
  const hash = crypto.createHash('md5').setEncoding('hex')
  return pipeline(readStream, hash).then(() => {
    return hash.read()
  })
}
```

app/js/S3Persistor.js
```diff
@@ -56,20 +67,73 @@ async function sendFile(bucketName, key, fsPath) {
...
+      hashPromise = new Promise((resolve, reject) => {
+        readStream.on('end', () => {
+          hash.end()
+          resolve(hash.read())
+        })
+        readStream.on('error', err => {
+          reject(err)
+        })
+      })
+    }
```
briangough: As for the other hash, I think this could be simplified to 

```
      const hash = crypto.createHash('md5').setEncoding('hex')
      hashPromise = pipeline(readStream, hash).then(() => {
        return hash.read()
      })
```

gh2k:
Great catch! ^_^ Thanks for checking so thoroughly.

It looks like the streams you get back from S3 start paused (as expected) but don't automatically resume when you add a `data` listener, as they should. This only affects the tests, as we use pipes everywhere in the main code, but the tests have a method for reading a string directly from the persistor to verify it, which uses a listener.

Calling `resume` on the stream fixes the problem. This seems to fix it: https://github.com/overleaf/filestore/pull/78/commits/9822ea6ab6d3ab284953cb406ce0252c48cd66be

briangough:
One other comment was `.on` + `removeListener` can be simplified to `.once`.

gh2k:
Thanks. Have tidied up in the refactor PR, here: https://github.com/overleaf/filestore/pull/79/commits/260d0add9758e9b3f27e94860f789c8a51cc9765

gh2k:
Merged #79 into here so that we have one fewer branch to worry about. Will treat this like an integration branch until the GCS stuff is ready to go

gh2k:
As mentioned in https://github.com/overleaf/filestore/pull/86#issuecomment-585908058 - it should be fine to deploy this, so will do rather than having extra branches around

PR #79, opened 2020-01-29
Refactor persistors to use a helper for common things

gh2k:
### Description

This adds a `PersistorHelper` (bit of a mouthful but I couldn't think of a better name for it) which contains some common functionality for the current persistors, and the new GCS persistor when it is ready.

Currently based onto `spd-migration-persistor` until that is merged.

#### Related Issues / PRs

Part of the work for overleaf/google-ops#731

### Review

Currently only `_wrapError` is shared by `FSPersistor` and `S3Perisstor`, but the other functionality here is useful to `GCSPersistor` as well.

I think it also makes a couple of mechanisms clearer - for example, the `waitForStreamReady` mechanism was bundled up as part of the general download logic and putting it in its own method defines what it's trying to do a bit more obviously.

There aren't any unit tests for `PersistorHelper`. Instead, I opted to test the functionality here via the main persistors instead.

#### Potential Impact

It touches the filestore streams, so high.

### DIFF
--- a/app/js/FSPersistor.js
+++ b/app/js/FSPersistor.js
@@ -8,6 +8,7 @@ const { promisify, callbackify } = require('util')
 
 const LocalFileWriter = require('./LocalFileWriter').promises
 const { NotFoundError, ReadError, WriteError } = require('./Errors')
+const PersistorHelper = require('./PersistorHelper')
 
 const pipeline = promisify(Stream.pipeline)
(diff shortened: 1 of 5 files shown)


gh2k:
Going to merge this into the `migration-persistor` branch so that there is one less branch to worry about. Don't want to deploy this until we've tested thoroughly on staging with the gcs persistor.

PR #80, opened 2020-02-03
[misc] cleanup install deps

das7pad:
### Description
I happen to have to rebuild the docker image again and it took ages. A quick inspection showed that you are still building an ancient version of ghostscript from source. We are using the debian package for imagemagick, so it should be fine to use their ghostscript package too. It is likely that they retain compatibility between the two, would on jump out of the line.

Program failures during the install_deps stage are no longer discarded, which reviled that the data dirs are included into the docker image - this was likely the trigger for the rebuild mentioned earlier. The data dirs are excluded from the COPY stage.

#### Potential Impact

Low, ghostscript is used for pdf to image conversion, which we do not allow in production.

#### Manual Testing Performed

- _[There is an acceptance test for the conversion](https://github.com/overleaf/filestore/blob/c8f869293273af3e6251f8d7bab714cea388cf96/test/acceptance/js/FilestoreTests.js#L335-L384) which passed._

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -5,3 +5,7 @@ gitrev
 .npm
 .nvmrc
 nodemon.json
+
+uploads/
+user_files/
+template_files/
--- a/install_deps.sh
+++ b/install_deps.sh
@@ -1,15 +1,14 @@
 #!/bin/sh
+
+set -ex
+
 apt-get update
 
-apt-get install vim imagemagick optipng --yes
+apt-get install ghostscript imagemagick optipng --yes
 
-wget -q https://s3.amazonaws.com/sl-public-dev-assets/ghostscript-9.15.tar.gz -O /tmp/ghostscript-9.15.tar.gz
-cd /tmp
-tar -xvf /tmp/ghostscript-9.15.tar.gz
-cd /tmp/ghostscript-9.15 && ./configure && make && make install
-npm rebuild
+rm -rf /var/lib/apt/lists/*
+
 mkdir /app/user_files/ /app/uploads/ /app/template_files/
 chown -R node:node /app/user_files
 chown -R node:node /app/uploads
 chown -R node:node /app/template_files
-ls -al /app


jdleesmiller:
CC @ShaneKilkelly @mserranom since this only really affects CE/SP at the moment.

.dockerignore
```diff
@@ -5,3 +5,7 @@ gitrev
 .npm
 .nvmrc
 nodemon.json
+
+uploads/
```
gh2k: :+1:

install_deps.sh
```diff
@@ -1,15 +1,14 @@
 #!/bin/sh
+
+set -ex
```
gh2k: :+1:

install_deps.sh
```diff
@@ -1,15 +1,14 @@
...
+
 apt-get update
 
-apt-get install vim imagemagick optipng --yes
+apt-get install ghostscript imagemagick optipng --yes
+
+rm -rf /var/lib/apt/lists/*
 
-wget -q https://s3.amazonaws.com/sl-public-dev-assets/ghostscript-9.15.tar.gz -O /tmp/ghostscript-9.15.tar.gz
```
gh2k: IDK what the custom build of Ghostscript was for, or whether it's still necessary. I *suspect* that we needed a newer version that didn't have a Debian pakage, but I don't know for sure. cc @ShaneKilkelly @henryoswald

install_deps.sh
```diff
@@ -1,15 +1,14 @@
...
-cd /tmp
-tar -xvf /tmp/ghostscript-9.15.tar.gz
-cd /tmp/ghostscript-9.15 && ./configure && make && make install
```
gh2k: I always feel like it's better to *not* have a C compiler if you don't need one. Are there dev tools we can remove?

das7pad: The `node:x.x.x` images are based on `buildpack-deps:DEBIAN_RELEASE`, which [adds a bunch of build tools](https://github.com/docker-library/buildpack-deps/blob/51a2871c1df6a6215227c1baeba81e15f04c6cff/stretch/Dockerfile) (and git + gpg tools + curl) to the respective debian image.

das7pad: We could test the `node:x.x.x-slim` variant for the second stage which is pretty much bare debian (-man pages, -docs) plus node.js: https://github.com/nodejs/docker-node/blob/46f5203674c748b0701135c8eeea4b250b3ecb6d/10/stretch-slim/Dockerfile

I am not familiar with the node-gyp process in regards to c-implementations/bindings. And how packages with other dependencies - e.g. for sqlite - behave when the libs are missing at runtime.

install_deps.sh
jdleesmiller: FWIW I'm not aware of any need for a custom build. I am fairly certain that it is not still necessary in the SaaS edition, because the feature that used ghostscript is presently disabled. (If you uploaded a .pdf to your project, the filestore would use ghostscript to generate a preview as a PNG/JPG that would work in-browser. We disabled it due to security concerns, but it's still available as an option in CE/SP builds.)

install_deps.sh
jdleesmiller: FWIW, I've not had any trouble with node-gyp and the `slim` images to date --- I think building on regular `node` and then running on `slim` is one of the main use cases for the `slim` builds.

install_deps.sh
das7pad: The [templates service uses that preview feature too](https://github.com/overleaf/templates-api/blob/14bd491f5fc9c05f1de8a6a855dfa0f8378e2492/app/coffee/filestore/TemplateBuilder.coffee#L96-L97), but I am not sure if its still in use.

jdleesmiller: That is a good point. It's no longer used in SaaS (we're not publishing any more SL templates) but again I think it would be used in SP.

das7pad: cc @mserranom

mserranom: no `install_deps.sh` is run in CE/ServerPro, this change shouldn't have any effect (and `ghostscript` is not installed in the image)

henryoswald: We had a custom build because ghostscript latest release at the time wasn't very up to date and head had some args we needed to use. If I remember correctly the Ghostscript dev's actually added them in for us very quickly.

install_deps.sh
das7pad: I'll have a look at moving to node-slim after the socket.io upgrade then.

das7pad: Let's merge this now and create a new ticket for node-slim.

das7pad: @gh2k agreed on this. I have raised https://github.com/overleaf/issues/issues/2724.

PR #82, opened 2020-02-12
Fix log info assignment in error handler

gh2k:
### Description

The line in the error handler to assign the error variable to the request handler was incorrect.

### Review

The handler is static, and so `this._logInfo` does not make any sense here. We should use the `requestLogger` object on the request instead.


gh2k:
> It might be worth to move `this` class into the logger module and add unit tests.

I think we did discuss doing this after decaf. :+1:

PR #86, opened 2020-02-13
Add GCS Persistor

gh2k:
### Description

Add a persistor to talk to GCS instead of S3.

~Targeted onto #78~ (#78 was merged)

#### Related Issues / PRs

Finishes off the work for overleaf/google-ops#731

### Review

It's a big chunk of new code I'm afraid. It's very similar to the S3 stuff obviously, but using the official GCS client.

This has only been tested as far as unit/acceptance tests go. Next steps should be to:

- Get this working with the dev env, using `fake-gcs-server` and migrating from FS
- Test on staging

I imagine both of these will show problems, but as the GCS stuff has only been tested against the fake server, "test on staging" is likely to be the more difficult.

#### Manual Testing Performed

- [ ] Just the tests, currently

### Deployment

We'll need to change the settings to use GCS and fall back to S3, once we're ready and have the necessary buckets and permissions in place.

The service account for the pods running this should have write permissions to the buckets. This will then get picked up automatically.

#### Deployment Checklist

- [ ] Create buckets
- [ ] Create a service account
- [ ] Add the service account to the filestore config
- [ ] Give the service account read/write permissions on the buckets
- [ ] Change the settings to use GCS as primary and fall back to the current S3 settings
- [ ] Test for a bit
- [ ] Enable fallback copies when we're sure it's working
- [ ] Start copying all of the data from the old S3 buckets so that we can eventually delete them

#### Metrics and Monitoring

- `gcs.egress` and `gcs.ingress` should be compared to `s3.ingress` and `s3.egress`
    - we should immediately see an end to `s3.egress`, which should get picked up by `gcs.egress`

#### Who Needs to Know?

We should definitely tell Support when this happens.

### DIFF
--- a/app/js/FileHandler.js
+++ b/app/js/FileHandler.js
@@ -10,6 +10,7 @@ const { ConversionError, WriteError } = require('./Errors')
 module.exports = {
   insertFile: callbackify(insertFile),
   deleteFile: callbackify(deleteFile),
+  deleteProject: callbackify(deleteProject),
   getFile: callbackify(getFile),
   getFileSize: callbackify(getFileSize),
   getDirectorySize: callbackify(getDirectorySize),
@@ -17,6 +18,7 @@ module.exports = {
     getFile,
     insertFile,
     deleteFile,
+    deleteProject,
     getFileSize,
     getDirectorySize
   }
@@ -46,6 +48,16 @@ async function deleteFile(bucket, key) {
     PersistorManager.promises.deleteFile(bucket, key),
     PersistorManager.promises.deleteDirectory(bucket, convertedKey)
   ])
+}
+
+async function deleteProject(bucket, key) {
+  if (!key.match(/^[0-9a-f]{24}\//i)) {
+    throw new WriteError({
+      message: 'key does not match validation regex',
+      info: { bucket, key }
+    })
+  }
+  await PersistorManager.promises.deleteDirectory(bucket, key)
 }
 
 async function getFile(bucket, key, opts) {
--- a/app/js/FileHandler.js
+++ b/app/js/FileHandler.js
@@ -5,7 +5,7 @@ const LocalFileWriter = require('./LocalFileWriter')
 const FileConverter = require('./FileConverter')
 const KeyBuilder = require('./KeyBuilder')
 const ImageOptimiser = require('./ImageOptimiser')
-const { ConversionError, WriteError } = require('./Errors')
+const { ConversionError, InvalidParametersError } = require('./Errors')
 
 module.exports = {
   insertFile: callbackify(insertFile),
@@ -27,7 +27,7 @@ module.exports = {
 async function insertFile(bucket, key, stream) {
   const convertedKey = KeyBuilder.getConvertedFolderKey(key)
   if (!convertedKey.match(/^[0-9a-f]{24}\/[0-9a-f]{24}/i)) {
-    throw new WriteError({
+    throw new InvalidParametersError({
       message: 'key does not match validation regex',
       info: { bucket, key, convertedKey }
     })
@@ -39,7 +39,7 @@ async function insertFile(bucket, key, stream) {
 async function deleteFile(bucket, key) {
   const convertedKey = KeyBuilder.getConvertedFolderKey(key)
   if (!convertedKey.match(/^[0-9a-f]{24}\/[0-9a-f]{24}/i)) {
-    throw new WriteError({
+    throw new InvalidParametersError({
       message: 'key does not match validation regex',
       info: { bucket, key, convertedKey }
     })
@@ -52,7 +52,7 @@ async function deleteFile(bucket, key) {
 
 async function deleteProject(bucket, key) {
   if (!key.match(/^[0-9a-f]{24}\//i)) {
-    throw new WriteError({
+    throw new InvalidParametersError({
       message: 'key does not match validation regex',
       info: { bucket, key }
     })
--- a/app/js/FileHandler.js
+++ b/app/js/FileHandler.js
@@ -5,7 +5,7 @@ const LocalFileWriter = require('./LocalFileWriter')
 const FileConverter = require('./FileConverter')
 const KeyBuilder = require('./KeyBuilder')
 const ImageOptimiser = require('./ImageOptimiser')
-const { ConversionError } = require('./Errors')
+const { ConversionError, WriteError } = require('./Errors')
 
 module.exports = {
   insertFile: callbackify(insertFile),
@@ -24,12 +24,24 @@ module.exports = {
 
 async function insertFile(bucket, key, stream) {
   const convertedKey = KeyBuilder.getConvertedFolderKey(key)
+  if (!convertedKey.match(/^[0-9a-f]{24}\/[0-9a-f]{24}/i)) {
+    throw new WriteError({
+      message: 'key does not match validation regex',
+      info: { bucket, key, convertedKey }
+    })
+  }
   await PersistorManager.promises.deleteDirectory(bucket, convertedKey)
   await PersistorManager.promises.sendStream(bucket, key, stream)
 }
 
 async function deleteFile(bucket, key) {
   const convertedKey = KeyBuilder.getConvertedFolderKey(key)
+  if (!convertedKey.match(/^[0-9a-f]{24}\/[0-9a-f]{24}/i)) {
+    throw new WriteError({
+      message: 'key does not match validation regex',
+      info: { bucket, key, convertedKey }
+    })
+  }
   await Promise.all([
     PersistorManager.promises.deleteFile(bucket, key),
     PersistorManager.promises.deleteDirectory(bucket, convertedKey)
--- a/app/js/GcsPersistor.js
+++ b/app/js/GcsPersistor.js
@@ -53,7 +53,7 @@ const GcsPersistor = {
 
 module.exports = GcsPersistor
 
-async function sendFile(bucket, key, fsPath) {
+async function sendFile(bucketName, key, fsPath) {
   let readStream
   try {
     readStream = fs.createReadStream(fsPath)
@@ -61,14 +61,14 @@ async function sendFile(bucket, key, fsPath) {
     throw PersistorHelper.wrapError(
       err,
       'error reading file from disk',
-      { bucketName: bucket, key, fsPath },
+      { bucketName, key, fsPath },
       ReadError
     )
   }
-  return sendStream(bucket, key, readStream)
+  return sendStream(bucketName, key, readStream)
 }
 
-async function sendStream(bucket, key, readStream, sourceMd5) {
+async function sendStream(bucketName, key, readStream, sourceMd5) {
   try {
     let hashPromise
 
@@ -96,7 +96,7 @@ async function sendStream(bucket, key, readStream, sourceMd5) {
     }
 
     const uploadStream = storage
-      .bucket(bucket)
+      .bucket(bucketName)
       .file(key)
       .createWriteStream(writeOptions)
 
@@ -107,25 +107,25 @@ async function sendStream(bucket, key, readStream, sourceMd5) {
     if (hashPromise) {
       sourceMd5 = await hashPromise
       // throws on mismatch
-      await PersistorHelper.verifyMd5(GcsPersistor, bucket, key, sourceMd5)
+      await PersistorHelper.verifyMd5(GcsPersistor, bucketName, key, sourceMd5)
     }
   } catch (err) {
     throw PersistorHelper.wrapError(
       err,
       'upload to GCS failed',
-      { bucket, key },
+      { bucketName, key },
       WriteError
     )
   }
 }
 
-async function getFileStream(bucket, key, opts = {}) {
+async function getFileStream(bucketName, key, opts = {}) {
   if (opts.end) {
     // S3 (and http range headers) treat 'end' as inclusive, so increase this by 1
     opts.end++
   }
   const stream = storage
-    .bucket(bucket)
+    .bucket(bucketName)
     .file(key)
     .createReadStream(opts)
 
@@ -141,16 +141,16 @@ async function getFileStream(bucket, key, opts = {}) {
     throw PersistorHelper.wrapError(
       err,
       'error reading file from GCS',
-      { bucket, key, opts },
+      { bucketName, key, opts },
       ReadError
     )
   }
 }
 
-async function getFileSize(bucket, key) {
+async function getFileSize(bucketName, key) {
   try {
     const metadata = await storage
-      .bucket(bucket)
+      .bucket(bucketName)
       .file(key)
       .getMetadata()
     return metadata[0].size
@@ -158,16 +158,16 @@ async function getFileSize(bucket, key) {
     throw PersistorHelper.wrapError(
       err,
       'error getting size of GCS object',
-      { bucket, key },
+      { bucketName, key },
       ReadError
     )
   }
 }
 
-async function getFileMd5Hash(bucket, key) {
+async function getFileMd5Hash(bucketName, key) {
   try {
     const metadata = await storage
-      .bucket(bucket)
+      .bucket(bucketName)
       .file(key)
       .getMetadata()
     return PersistorHelper.base64ToHex(metadata[0].md5Hash)
@@ -175,23 +175,23 @@ async function getFileMd5Hash(bucket, key) {
     throw PersistorHelper.wrapError(
       err,
       'error getting hash of GCS object',
-      { bucket, key },
+      { bucketName, key },
       ReadError
     )
   }
 }
 
-async function deleteFile(bucket, key) {
+async function deleteFile(bucketName, key) {
   try {
     await storage
-      .bucket(bucket)
+      .bucket(bucketName)
       .file(key)
       .delete()
   } catch (err) {
     const error = PersistorHelper.wrapError(
       err,
       'error deleting GCS object',
-      { bucket, key },
+      { bucketName, key },
       WriteError
     )
     if (!(error instanceof NotFoundError)) {
@@ -200,17 +200,19 @@ async function deleteFile(bucket, key) {
   }
 }
 
-async function deleteDirectory(bucket, key) {
+async function deleteDirectory(bucketName, key) {
   let files
 
   try {
-    const response = await storage.bucket(bucket).getFiles({ directory: key })
+    const response = await storage
+      .bucket(bucketName)
+      .getFiles({ directory: key })
     files = response[0]
   } catch (err) {
     const error = PersistorHelper.wrapError(
       err,
       'failed to list objects in GCS',
-      { bucket, key },
+      { bucketName, key },
       ReadError
     )
     if (error instanceof NotFoundError) {
@@ -226,7 +228,7 @@ async function deleteDirectory(bucket, key) {
       const error = PersistorHelper.wrapError(
         err,
         'failed to delete object in GCS',
-        { bucket, key },
+        { bucketName, key },
         WriteError
       )
       if (!(error instanceof NotFoundError)) {
@@ -236,17 +238,19 @@ async function deleteDirectory(bucket, key) {
   }
 }
 
-async function directorySize(bucket, key) {
+async function directorySize(bucketName, key) {
   let files
 
   try {
-    const response = await storage.bucket(bucket).getFiles({ directory: key })
+    const response = await storage
+      .bucket(bucketName)
+      .getFiles({ directory: key })
     files = response[0]
   } catch (err) {
     throw PersistorHelper.wrapError(
       err,
       'failed to list objects in GCS',
-      { bucket, key },
+      { bucketName, key },
       ReadError
     )
   }
@@ -254,10 +258,10 @@ async function directorySize(bucket, key) {
   return files.reduce((acc, file) => Number(file.metadata.size) + acc, 0)
 }
 
-async function checkIfFileExists(bucket, key) {
+async function checkIfFileExists(bucketName, key) {
   try {
     const response = await storage
-      .bucket(bucket)
+      .bucket(bucketName)
       .file(key)
       .exists()
     return response[0]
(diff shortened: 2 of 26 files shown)


jdleesmiller:
Would it be feasible to deploy https://github.com/overleaf/filestore/pull/78 first, just without the migration persister configured? I'm thinking we can reduce our overall risk by deploying this in some smaller units --- we'd expect #78 without the persister to preserve current behaviour, right?

gh2k:
Yes, good point. I'm a fan of a good integration branch, but you're right that it shouldn't harm anything deploying the updated code. - I think most of it won't get executed because `PersistorManager` won't pull it in, so I don't think it makes a huge difference... But yes it does mitigate some risk and result in fewer branches hanging around. Will deploy. :+1:

test/acceptance/deps/Dockerfile.fake-gcs
```diff
@@ -0,0 +1,5 @@
+FROM fsouza/fake-gcs-server
+RUN apk add --update --no-cache curl
+COPY healthcheck.sh /healthcheck.sh
+HEALTHCHECK --interval=1s --timeout=1s --retries=30 CMD /healthcheck.sh http://localhost:9090
+CMD ["--port=9090"]
```
das7pad:
```suggestion
CMD ["--port=9090","--scheme=http"]
```
```
# terminal 1
$ sudo docker run --rm --name fake-gcs fsouza/fake-gcs-server --port 9090 --scheme http 
time="2020-02-14T16:07:01Z" level=info msg="couldn't load any objects or buckets from \"/data\", starting empty"
time="2020-02-14T16:07:01Z" level=info msg="server started at http://[::]:9090"
time="2020-02-14T16:07:42Z" level=info msg="172.17.0.1 - - [14/Feb/2020:16:07:42 +0000] \"GET /storage/v1/b HTTP/1.1\" 200 54"

# terminal 2
$ curl `sudo docker inspect -f '{{.NetworkSettings.IPAddress}}' fake-gcs`:9090/storage/v1/b -i
HTTP/1.1 200 OK
Date: Fri, 14 Feb 2020 16:07:42 GMT
Content-Length: 54
Content-Type: text/plain; charset=utf-8

{"kind":"storage#buckets","items":[],"prefixes":null}
```

docker-compose.ci.yml
```diff
@@ -22,20 +22,27 @@ services:
...
       AWS_S3_PUBLIC_FILES_BUCKET_NAME: fake_public_files
+      AWS_S3_ENDPOINT: http://s3:9090
+      AWS_ACCESS_KEY_ID: fake
+      AWS_SECRET_ACCESS_KEY: fake
+      GCS_API_ENDPOINT: gcs:9090
+      GCS_USER_FILES_BUCKET_NAME: fake_userfiles
+      GCS_TEMPLATE_FILES_BUCKET_NAME: fake_templatefiles
+      GCS_PUBLIC_FILES_BUCKET_NAME: fake_publicfiles
+      NODE_TLS_REJECT_UNAUTHORIZED: 0
+      STORAGE_EMULATOR_HOST: https://gcs:9090/storage/v1
```
das7pad: Let's not mess with tls auth and simply use the `http` scheme, see changes in the `Dockerfile`.

app/js/GcsPersistor.js
```diff
@@ -0,0 +1,291 @@
...
+      { bucket, key },
+      ReadError
+    )
+    if (error instanceof NotFoundError) {
+      return
+    }
+    throw error
+  }
+
+  for (const index in files) {
```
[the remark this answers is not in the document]
briangough: I think we should use the [Bucket#deleteFiles](https://googleapis.dev/nodejs/storage/latest/Bucket.html#deleteFiles) method provided by Google here (with `force:true`).  It is not atomic but it does allow deleting all files in a bucket under a `prefix/` with a single call. 

It would be a good idea to check that the prefix has the expected form so that we cannot delete the whole bucket by mistake!

app/js/GcsPersistor.js
```diff
@@ -0,0 +1,291 @@
...
+    directorySize
+  }
+}
+
+module.exports = GcsPersistor
+
+async function sendFile(bucket, key, fsPath) {
+  let readStream
+  try {
+    readStream = fs.createReadStream(fsPath)
```
briangough: I don't think `fs.createReadStream` throws any exceptions, but does emit error events.

I think what is going to happen here is that any error event on the readStream is going to be returned in the callback of the metered stream, and that looks like it will be discarded.

app/js/GcsPersistor.js
```diff
@@ -0,0 +1,291 @@
...
+
+    // if there is no supplied md5 hash, we calculate the hash as the data passes through
+    if (!sourceMd5) {
+      hashPromise = PersistorHelper.calculateStreamMd5(readStream)
+    }
+
+    const meteredStream = PersistorHelper.getMeteredStream(
+      readStream,
+      (_, byteCount) => {
+        metrics.count('gcs.egress', byteCount)
```
briangough: As mentioned above any error from the readstream piping into the metered stream is ignored here.

app/js/GcsPersistor.js
```diff
@@ -0,0 +1,291 @@
...
+        md5Hash: PersistorHelper.hexToBase64(sourceMd5)
+      }
+    }
+
+    const uploadStream = storage
+      .bucket(bucket)
+      .file(key)
+      .createWriteStream(writeOptions)
+
+    await pipeline(meteredStream, uploadStream)
```
briangough: Errors from readStream won't be forwarded to meteredStream, so we will lose them.  Can we just use readStream here?

Given that we'll have bigger problems if  we see significant numbers of errors in actual usage, could we remove this metering of streams (in the name of simpler code) and just increment the metric by the number of bytes on successful upload?  Our error rate must be pretty low.

app/js/GcsPersistor.js
```diff
@@ -0,0 +1,291 @@
...
+    opts.end++
+  }
+  const stream = storage
+    .bucket(bucket)
+    .file(key)
+    .createReadStream(opts)
+
+  const meteredStream = PersistorHelper.getMeteredStream(stream, (_, bytes) => {
+    // ignore the error parameter and just log the byte count
+    metrics.count('gcs.ingress', bytes)
```
briangough: I find the ingress and egress metric names a bit confusing, e.g.  `gcs.ingress` is not ingress to gcs, but ingress to our cluster?  Would `download` and `upload` be clearer?

app/js/GcsPersistor.js
```diff
@@ -0,0 +1,291 @@
...
+  })
+
+  try {
+    await PersistorHelper.waitForStreamReady(stream)
+    return meteredStream
```
briangough: I think there is an issue again of errors on the `stream` not being forwarded to `meteredStream`.

app/js/GcsPersistor.js
briangough: I just saw the comment 

> gcs.egress and gcs.ingress should be compared to s3.ingress and s3.egress
> we should immediately see an end to s3.egress, which should get picked up by gcs.egress

Better to stick with this as we are already using these metric names.  Maybe a comment about it in the code as I had to think twice what they meant.

gh2k:
@briangough 
```JavaScript
const meter = require('stream-meter')
const fs = require('fs')
const Stream = require('stream')
const { promisify } = require('util')

const pipeline = promisify(Stream.pipeline)

const readStream = fs.createReadStream('/somefilethatdoesntexist')
const meteredStream = meter()

pipeline(readStream, meteredStream).catch(err => console.log('there is an expected error here, but will we get one later?'))

writeStream = fs.createWriteStream('/tmp/foo')
pipeline(meteredStream, writeStream).catch(err => console.log('we got a later error'))
```
yields
```
-> node test.js
there is an expected error here, but will we get one later?
we got a later error
```
WDYT?

gh2k:
Also:
```JavaScript
const meter = require('stream-meter')
const fs = require('fs')
const Stream = require('stream')
const { promisify } = require('util')

const pipeline = promisify(Stream.pipeline)

const readStream = fs.createReadStream('/somefilethatdoesntexist')
const meteredStream = meter()

writeStream = fs.createWriteStream('/tmp/foo')
pipeline(meteredStream, writeStream).catch(err => console.log('so, will we get two errors this time'))

pipeline(readStream, meteredStream).catch(err => console.log('will we?'))
```
yields
```
-> node test.js 
will we?
so, will we get two errors this time
```

So, I think `pipeline` is smart enough to handle this...

briangough:
You are right, you can ignore my comments about it not being forwarded. Here is the relevant part of the documentation:

> A module method to pipe between streams _forwarding errors_ and properly cleaning up and provide a callback when the pipeline is complete.

The following part has been added in the node 12 documentation (it wasn't there before).


> stream.pipeline() will call stream.destroy(err) on all streams except:
> 
> Readable streams which have emitted 'end' or 'close'.
> Writable streams which have emitted 'finish' or 'close'.

https://nodejs.org/api/stream.html#stream_stream_pipeline_streams_callback

app/js/GcsPersistor.js
gh2k: I've always read `egress` and `ingress` as from the perspective of the related service. "egress from me", "ingress to me". As opposed to `upload` and `download` which can be more easily confused imho.

But, it gets more confusing if you're paying for both ends I guess. ;-) -- I'll add a comment.

test/unit/js/GcsPersistorTests.js
```diff
@@ -0,0 +1,756 @@
...
+      })
+
+      it('should generate a WriteError', function() {
+        expect(error).to.be.an.instanceOf(Errors.WriteError)
+      })
+
+      it('should wrap the error', function() {
+        expect(error.cause).to.equal(genericError)
+      })
+    })
```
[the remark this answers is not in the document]
gh2k: Will use Bucket.deleteFiles instead

app/js/GcsPersistor.js
```diff
@@ -0,0 +1,291 @@
...
+  }
+
+  for (const index in files) {
+    try {
+      await files[index].delete()
+    } catch (err) {
+      const error = PersistorHelper.wrapError(
+        err,
+        'failed to delete object in GCS',
+        { bucket, key },
```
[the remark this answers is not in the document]
gh2k: Will use Bucket.deleteFiles instead

gh2k:
> It passes in CI so it must be something I'm doing but I'm not sure what. Any ideas?
```
cd filestore/
docker-compose build --pull --no-cache
```
?

gh2k:
Added overleaf/dev-environment#318 to get this working in the dev env.

briangough:
> cd filestore/
> docker-compose build --pull --no-cache

Thanks, that fixed it.

app/js/GcsPersistor.js
```diff
@@ -0,0 +1,273 @@
...
+      WriteError
+    )
+    if (!(error instanceof NotFoundError)) {
+      throw error
+    }
+  }
+}
+
+async function deleteDirectory(bucketName, key) {
+  if (!key.match(/^[a-z0-9_-]+/i)) {
```
briangough: For safety, how about we prevent deleting anything more than the content for a specific file:
```
!key.match(/^[0-9a-f]{24}\/[0-9a-f]{24}/i)
```

gh2k: yes, that is better

gh2k: So, when running the acceptance tests I'm seeing URLs such as
```
acceptance_tests_0.9991905513266544/directory/0.23709030119276875-converted-cache/
```
which don't match.

One thing that occurs is that the persistor is somewhat generic, so enforcing this path style (objectid/objectid) here where we're not doing it in other places is a bit awkward. For example, if I want to plug this in to some other service like docstore, where we want to migrate from S3 to GCS, the logic then gets in the way.

briangough: Replacing the `id = Math.random()`'s in the acceptance tests with `ObjectId()` (which creates a random objectid) should fix most of those failures.

I think it would be easy enough to make this check configurable if we need to use the code elsewhere.  

I do think it's worth having multiple layers of defenses against accidentally deleting our entire filestore bucket as this would not be easy to recover from!

gh2k:
Added a commit to handle the bucket config that we would like to use. See https://github.com/overleaf/google-ops/pull/893 for the details on how we want to set up the buckets. But, basically on delete we want to:
* clear an event-based hold that will be on the file (so that it can be deleted)
* copy the file to another bucket (where it will be deleted later)
* delete the file

gh2k:
> I do not think that this scales properly

You have a point. Currently, we're not deleting much (anything?) but we're hoping that will change. The most efficient way to delete a project would be to expose `deleteDirectory` as an endpoint (e.g. `DELETE /project/{id}` and have filestore do it, rather than iterate from web.

Some kind of async-limit would help here. It would still as you say requires 3k calls into GCS, but at least we could do lots of them at once.

@jdleesmiller had another idea of having just one bucket with a retention period and: 
- a default event-based lock
- a retention period (which starts when the lock is removed)
- a lifecycle rule (which deletes the file at the end of the retention period)

The issue I have with this is that a request to the file doesn't look like a 404 once it has been deleted. However *both* implementations have a problem with overwriting files.

With a single bucket, you can't overwrite a file until its retention period is up. With two buckets, as-implemented, we have kicked that can down the road a little in that you can't delete a file a second time because it won't be able to overwrite the previously-deleted one.

In the latter case, we can work around this issue in principle because we can archive the file off to the 2nd bucket with a timestamp appended to its name. We'd only ever be recovering these files manually anyway.

WRT deleting 2k files with 3 sequential ops in parallel (with a limit?) - I'm not sure how long this will take and how much of a problem it would be. According to Mongo, an average of ~2901.5 documents are deleted per day, so I don't think that concurrent-deletes-of-many-projects is going to be a huge problem. - In fact, we have an average of 1.632 files per deleted project - But it's worth seeing how long it takes to do 2,000 3-op deletions in parallel to see if this is likely to cause us an issue when deleting a project of max size.

I'll refactor the code here to do that and give it a try.

gh2k:
Have (conditionally) enabled the above behaviour and added a project-deletion endpoint.

I move the logic for determining whether the directory key is valid out to `FileHandler` instead of in the persistor. This way we can keep it specific to filestore whilst keeping the persistor as reusable as possible, without relying on settings.

Modified `FSPersistor.deleteDirectory` to use the same behaviour as S3 and GCS, in that it deletes everything starting with the supplied prefix instead of specifically looking for a directory with that name.

The jury, it seems, is still out on whether we need the `unlock-copy-delete` mechanism. Google support are confused about whether or not the lifecycle events work as we would like (e.g. lifecycle timer starts at deletion time, as opposed to at upload time) and are looking into it. I'm also doing my own tests to try and confirm what happens here.

`unlock-copy-delete` is still safer WRT accidental deletion in that it's not possible to use the CLI or web interface to delete files accidentally, although it could instead be replaced by `unlock-delete`. (Either or neither of these behaviours are configurable in this PR, now, but let's wait and see what Google and our tests say.)

das7pad:
> With a single bucket, you can't overwrite a file until its retention period is up.

As far as I can tell there is not way to overwrite a file from the UI. There are `delete` and `rename` actions per file only. So we would always end up with a new file id per revision (due to a delete op and new upload).

The event-based solutions looks promising then.

gh2k:
> As far as I can tell there is no way to overwrite a file from the UI.

Indeed. One thing I'm trying to balance with the `*Persistor`s is that we can feasibly use them for Other Things if we want to migrate other services from S3 or similar. So for that reason, I've been trying to keep them as generic as possible and keep the filestore-specific mechanisms in `FileController`/`FileHandler`.

app/js/GcsPersistor.js
```diff
@@ -0,0 +1,283 @@
...
+      'upload to GCS failed',
+      { bucketName, key },
+      WriteError
+    )
+  }
+}
+
+async function getFileStream(bucketName, key, opts = {}) {
+  if (opts.end) {
+    // S3 (and http range headers) treat 'end' as inclusive, so increase this by 1
```
briangough: would be useful to add a link to https://googleapis.dev/nodejs/storage/latest/global.html#CreateReadStreamOptions  for more explanation

we are modifying `opts.end` in-place. I don't think that matters here, but maybe it's safer to clone the options.

package.json
```diff
@@ -55,11 +57,13 @@
     "eslint-plugin-promise": "^4.2.1",
     "eslint-plugin-standard": "^4.0.1",
     "mocha": "5.2.0",
+    "mongodb": "^3.5.4",
```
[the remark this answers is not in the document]
gh2k: I think given the number of `ObjectId()` calls we have we might run the risk of running out of entropy, but yes something like this is probably better.

app/js/FileHandler.js
```diff
@@ -5,37 +5,61 @@ const LocalFileWriter = require('./LocalFileWriter')
...
 
 async function insertFile(bucket, key, stream) {
   const convertedKey = KeyBuilder.getConvertedFolderKey(key)
+  if (!convertedKey.match(/^[0-9a-f]{24}\/[0-9a-f]{24}/i)) {
+    throw new WriteError({
+      message: 'key does not match validation regex',
+      info: { bucket, key, convertedKey }
+    })
+  }
   await PersistorManager.promises.deleteDirectory(bucket, convertedKey)
```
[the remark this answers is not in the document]
das7pad: > 2\. remove the folder if someone uploads a file ending in `-converted-cache` (unlikely but we need to handle it)

Files are stored with the file-id as name. 
```
$ tree filestore/user_files/   
filestore/user_files/
|-- 5e2b0d6bacb2520037d92bd3_5e2efcd3b5378700256ad4d4
#   projectId                                  fileId
```

gh2k:
Also noting here that JLM suggests in https://github.com/overleaf/google-ops/pull/893 that we reverse the order of `unlock -> copy` to `copy -> unlock` on delete, which seems like a good idea.

app/js/FileHandler.js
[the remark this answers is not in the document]
gh2k: In the spirit of "keep the persistor code clean and reusable and the filestore-specific logic in the controllers/handlers", and in the interests of controlling the rolling snowball, I wonder if we could keep this as-is for now and revisit later.

Ideally I'd like to see (as you suggest) `FileHandler` implement its own cached-file cleanup mechanism by passing a different bucket into `PersistorManager.deleteDirectory`.

As @das7pad suggests, in practice we can only upload files in the form of ProjectId/FileId anyway, so this logic isn't in the way of anything here. I would rather not spend time tidying up conversions-related things at this point. WDYT?

app/js/FileHandler.js
```diff
@@ -5,37 +5,61 @@ const LocalFileWriter = require('./LocalFileWriter')
...
 }
 
 async function deleteFile(bucket, key) {
   const convertedKey = KeyBuilder.getConvertedFolderKey(key)
+  if (!convertedKey.match(/^[0-9a-f]{24}\/[0-9a-f]{24}/i)) {
+    throw new WriteError({
+      message: 'key does not match validation regex',
+      info: { bucket, key, convertedKey }
+    })
+  }
```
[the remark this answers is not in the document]
gh2k: See answer above

gh2k:
> But it's worth seeing how long it takes to do 2,000 3-op deletions in parallel to see if this is likely to cause us an issue when deleting a project of max size. I think if we can do that in a timely fashion then we'll be OK.

Project with 1995 jpeg files on Staging, with a concurrency of 50:
```
node@filestore-stag-9f874d8c5-z2zdh:/app$ time curl -X DELETE 'http://localhost:3009/project/5e7111d80a9cde0001979d03'

real	0m19.869s
user	0m0.007s
sys	0m0.005s
```

I guess if the files were 50MB each this would take longer due to all of the copies. It's worth considering the timeout when we eventually call into this endpoint from somewhere, but I think we can live with this. It takes a lot longer to upload the zip in the first place. ;-)

app/js/FileHandler.js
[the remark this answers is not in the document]
briangough: Yes fair point,  no need to change it now.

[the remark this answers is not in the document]
briangough: > Files are stored with the file-id as name.

Yes, my mistake.

package.json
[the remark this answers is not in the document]
briangough: True.  How about `crypto.createHash('md5').update(`${counter++}`).digest("hex").slice(0,24)`?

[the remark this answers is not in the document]
gh2k: Pinched this from SO which seems to work just fine
https://github.com/overleaf/filestore/pull/86/commits/f3d503cb0fbc62cebad198385ce5f2c010026d62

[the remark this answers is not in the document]
jdleesmiller: Thanks, I have learned some things about object IDs due to this discussion :)

While I am generally in favour of not having too many dependencies... I feel like having to jump through these hoops and/or worry about running out of entropy is probably enough of a benefit that we should keep the dependency. It's server-side only, and it's something we depend on all over, so I don't think it is particularly bad.

If we do the planned (or at least being considered?) work to ship only production dependencies instead of both dev and production dependencies, it would also affect only dev.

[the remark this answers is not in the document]
gh2k: Happy to drop the commit

gh2k:
For completeness:
```
node@filestore-stag-9f874d8c5-7p6jt:/app$ time curl -X DELETE http://localhost:3009/project/5e721311d70a120001ef4869

real	0m5.900s
user	0m0.007s
sys	0m0.004s
```
This project had 2,000 25MB files. I didn't believe it had done it, but the files are there in the deleted bucket and not there in the original one.

gh2k:
Force-pushed to remove the `dont-use-mongodb-for-ObjectId` commit.

The config for this is ready to go in `google-ops#master` and we're on staging with GCS -> S3.

Created https://github.com/overleaf/google-ops/pull/900 which reverts the config to use S3 as primary in case something goes wrong, but hopefully we won't need it.

Copy-on-miss is on in staging, but will be off in prod initially.

The production buckets and service accounts are deployed. I think we are good to go.

PR #90, opened 2020-02-14
Fix eTags not being in md5 format for large files

gh2k:
### Description

When uploading a file, we verify the md5 based on the eTag. By default, `upload` will use a multipart upload if the file is over 5MiB.

When using a multipart upload, the md5 hash is not correctly calcuated, and is returned in a different format.

This PR:

- sets the max chunk size to 100MiB, to ensure files are uploaded in a single part
- fetches the file to calculate the md5 in the case where the md5 is not in the correct format

The latter should fix issues where we need to calculate the md5 hash of a file stored in S3 which has been uploaded as a multipart upload. This shouldn't need to happen very often.

### Review

Without setting the chunk size, this could cause a lot of additional ingress as we would be re-downloading any >5MiB file. However, in pretty much every case we should be uploading in a single part, in which case the eTag will be the md5 hash.

#### Potential Impact

S3 uploads

#### Manual Testing Performed

- [ ] Upload a file bigger than 5MiB

#### Metrics and Monitoring

We should keep track of the `s3.ingress` and `s3.egress` metrics, to verify that this does not have a significant impact on bandwidth.

### DIFF
--- a/app/js/S3Persistor.js
+++ b/app/js/S3Persistor.js
@@ -98,7 +98,7 @@ async function sendStream(bucketName, key, readStream, sourceMd5) {
     const response = await _getClientForBucket(bucketName)
       .upload(uploadOptions, { partSize: 100 * 1024 * 1024 })
       .promise()
-    const destMd5 = _md5FromResponse(response)
+    const destMd5 = await _md5FromResponse(response)
 
     // if we didn't have an md5 hash, we should compare our computed one with S3's
     // as we couldn't tell S3 about it beforehand
@@ -219,7 +219,7 @@ async function getFileMd5Hash(bucketName, key) {
     const response = await _getClientForBucket(bucketName)
       .headObject({ Bucket: bucketName, Key: key })
       .promise()
-    const md5 = _md5FromResponse(response)
+    const md5 = await _md5FromResponse(response)
     return md5
   } catch (err) {
     throw PersistorHelper.wrapError(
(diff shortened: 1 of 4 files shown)


briangough:
> * sets the max chunk size to 100MiB, to ensure files are uploaded in a single part
> * fetches the file to calculate the md5 in the case where the md5 is not in the correct format
> 
> The latter should fix issues where we need to calculate the md5 hash of a file stored in S3 which has been uploaded as a multipart upload. This shouldn't need to happen very often.

Just to check my understanding, apart from checking the content of an uploaded file by downloading it as part of `sendStream`, what is the other case where we would need to fetch the file?

gh2k:
> I think it would be good to have an acceptance test which checks the large file case, if the fake s3 modules supports it.

Good call. I'll see if it does.

> we could also store the MD5/SHA1 as an object property

I think we only have a smallish window where we will be writing to S3 via this mechanism - i.e. the lag between the migration code being deployed and the GCS code being deployed, so I don't think that this is worth the effort. - Particularly as uploading the files in single chunks should work around the need for this anyway.

> Just to check my understanding, apart from checking the content of an uploaded file by downloading it as part of sendStream, what is the other case where we would need to fetch the file?

I've been thinking about this over the weekend. I think this may have introduced an issue where we end up downloading the entire file to check its MD5 before sending it to GCS, instead of computing it as we go - causing an additional download and delay which I don't think we need to do. We need to make sure this doesn't happen.

gh2k:
Added acceptance tests (the fake-s3 server *does* match the actual S3 behaviour here - it fails if the chunk size is too small) and removed the case where we would download the file a second time

PR #92, opened 2020-02-17
Add migration persistor with md5 fix

gh2k:
reopening of #78 with the md5 fix from #90


gh2k:
Conflicts are messed up. Will reopen later

PR #93, opened 2020-02-17
Add migration persistor with md5 fix

gh2k:
Redux of #92, rebased onto master properly, and reverting the revert commit.

Includes the original #78 and the md5 fix from #90


config/settings.defaults.coffee
```diff
@@ -38,6 +38,7 @@ settings =
 				key: process.env['AWS_ACCESS_KEY_ID']
 				secret: process.env['AWS_SECRET_ACCESS_KEY']
 				endpoint: process.env['AWS_S3_ENDPOINT']
+				partSize: process.env['S3_PARTSIZE'] or (100 * 1024 * 1024)
```
das7pad:
```suggestion
				partSize: process.env['AWS_S3_PARTSIZE'] or (100 * 1024 * 1024)
```
Let's be consistent with the other environment variables and prefix it with `AWS_`.

PR #94, opened 2020-02-27
Write to error log on unhandled rejection/exception

gh2k:
### Description

Clutching at straws a little bit, but this will write to the error log on unhandled promise rejections or unhandled exceptions, and exit with an error code on the latter.

Clutching at straws a bit, as I've so far been unable to get any insight into what is going wrong. This might help.

### DIFF
--- a/app.js
+++ b/app.js
@@ -156,4 +156,13 @@ if (!module.parent) {
   })
 }
 
+process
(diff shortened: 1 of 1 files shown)


henryoswald:
I think this is good to have by default in all apps anyway.

PR #95, opened 2020-02-28
[misc] test cleanup

das7pad:
### Description

These are all rather unrelated patches -- they all touch tests. So I recommend reviewing them on a per-commit basis. 

### Review

- [misc] test/unit: add missing require stubs for metrics and settings f0c8eac

  less verbose tests and slightly better performance due to omitted imports

- [misc] test/unit: add missing globals that are lazy loaded 2300727
  
  preparation for node 12

- [misc] test/unit: KeybuilderTests: use a unique suite label 3b7e33a

  fix copy and paste error. useful for limiting the number of running test suites via `MOCHA_GREP`

- [misc] test/acceptance: harden the startup check for s3 57edd4b

  minio -- my s3 replacement -- returns a 403 without auth, failing the current startup check. Placing a file in s3 as the startup check functions as an extended healthcheck too. Also limits the timeout to the one of the test-suite and returns a descriptive error just before hitting the timeout.

- [misc] test/acceptance: skip the shutdown in case we did not start yet 2e3cb96

  otherwise there are errors from cleanup when the startup fails.

- [misc] test/acceptance: do not hard code fake credentials 480470f

  pickup auth from env vars instead -- useful when there are actual credentials required, e.g. for minio.

- [misc] test/acceptance: retrieve ingress metrics just before using it d831a31

  The upload request can bump the ingress metric.
  The content hash validation might require a full download
  in case the ETag field of the upload response is not a md5 sum.

#### Potential Impact
Low, only test files are touched.

### DIFF
--- a/test/unit/js/S3PersistorTests.js
+++ b/test/unit/js/S3PersistorTests.js
@@ -158,7 +158,7 @@ describe('S3PersistorTests', function() {
         'metrics-sharelatex': Metrics,
         crypto
       },
-      globals: { console }
+      globals: { console, Buffer }
     })
   })
 
--- a/test/acceptance/js/FilestoreApp.js
+++ b/test/acceptance/js/FilestoreApp.js
@@ -4,11 +4,7 @@ const fs = require('fs')
 const Path = require('path')
 const { promisify } = require('util')
 const disrequire = require('disrequire')
-const rp = require('request-promise-native').defaults({
-  resolveWithFullResponse: true
-})
-
-const S3_TRIES = 30
+const AWS = require('aws-sdk')
 
 logger.logger.level('info')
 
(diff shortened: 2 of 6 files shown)


test/unit/js/S3PersistorTests.js
```diff
@@ -158,7 +158,7 @@ describe('S3PersistorTests', function() {
         'metrics-sharelatex': Metrics,
         crypto
       },
-      globals: { console }
+      globals: { console, Buffer }
```
gh2k: Out of curiosity, why do we need this in node 12? What happens if we don't add it?

das7pad: `Buffer` is not defined downstream then.

```
  1) S3PersistorTests
       sendStream
         when a hash is supploed
           "before each" hook for "should not calculate the md5 hash of the file":
     WriteError: upload to S3 failed: Buffer is not defined
      at Object.wrapError (app/js/PersistorHelper.js:100:12)
      at Object.sendStream (app/js/S3Persistor.js:126:27)
      at Context.<anonymous> (test/unit/js/S3PersistorTests.js:488:37)
      at processImmediate (internal/timers.js:439:21)
```

PR #96, opened 2020-02-28
[misc] dependencies cleanup

das7pad:
### Description

- fs.extra

  See the detailed commit message 83e35b2

- underscore

  Only one function `_.once` is used at this time. A simple counter works too 
e11f8f7
- heapdump

  As discussed in slack with @gh2k , the `/heapdump` route is not used anymore and can be removed.

- mocha -> dev

  This is in preparation of installing only production dependencies in production images.

- pngcrush

  Not used since 2014 https://github.com/overleaf/filestore/commit/a9aa696106fcbe0ad8d3c97d1fa9f01e061c4a05#diff-f284c7d92133cc92d9ff44c85c55ab63.

#### Potential Impact
Low


app/js/SafeExec.js
```diff
@@ -28,12 +27,14 @@ function safeExec(command, options, callback) {
 
   let killTimer
 
-  const cleanup = _.once(function(err) {
+  let cleanedUp = 0
```
jdleesmiller: I feel like this is less clear (for humans) than `_.once`, so while it is a bit overkill to load all of underscore for this one thing, I'm not sure we should sacrifice readability. What about using https://www.npmjs.com/package/lodash.once ?

das7pad: Agree. I rather put the logic in a local function, than using this 294 lines package. https://github.com/lodash/lodash/blob/4.1.1-npm-packages/lodash.once/index.js

app/js/SafeExec.js
```diff
@@ -87,3 +85,12 @@ function safeExecPromise(command, options) {
     })
   })
 }
+
+module.exports.runOnce = runOnce
+function runOnce(fn = function() {}) {
```
gh2k: Do we need the default implementation here? I feel like this function probably wouldn't be of much use without it.

app/js/SafeExec.js
```diff
@@ -87,3 +85,12 @@ function safeExecPromise(command, options) {
     })
   })
 }
+
+module.exports.runOnce = runOnce
```
gh2k: Does this need to be exported? I don't think it is used outside of this module.

app/js/SafeExec.js
das7pad: No, I'll remove it.

app/js/SafeExec.js
das7pad: I wrote a quick unit-tests for it and this test needs access to this internal fn.

https://github.com/overleaf/filestore/pull/96/files#diff-ecdb1e7e03113c4eee7b501b26f99999

I can move it to a private `_runOnce` export if you would like.

gh2k: No, that makes sense - leave it as-is.

app/js/SafeExec.js
jdleesmiller: That is perhaps a surprising amount of code... but TBH I'd rather have 294 lines of upstream code, complete with tests and documentation, etc., than 30 lines of code (and tests) and consequent discussion in this PR that we have to maintain. I would still prefer to use a library here.

das7pad: Done in https://github.com/overleaf/filestore/pull/96/commits/ecdaff14f0647e8a7e7ec9f1775b472749ef4ca7

PR #98, opened 2020-02-28
[misc] promisify FileHandler and remove dependency on async

das7pad:
### Description
Followup for #60

This PR promisifies the FileHandler.

It also removes a now unused dependency `async`.

#### Related Issues / PRs
#60

#### Potential Impact
High, all requests run through this handler.

### DIFF
--- a/app/js/FileHandler.js
+++ b/app/js/FileHandler.js
@@ -1,176 +1,153 @@
-const { promisify } = require('util')
+const { callbackify } = require('util')
 const fs = require('fs')
 const PersistorManager = require('./PersistorManager')
 const LocalFileWriter = require('./LocalFileWriter')
 const FileConverter = require('./FileConverter')
 const KeyBuilder = require('./KeyBuilder')
-const async = require('async')
 const ImageOptimiser = require('./ImageOptimiser')
 const { ConversionError } = require('./Errors')
 
 module.exports = {
-  insertFile,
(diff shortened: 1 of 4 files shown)


app/js/FileHandler.js
```diff
@@ -1,176 +1,153 @@
...
 
-    const done = function(err, destPath) {
-      if (err) {
-        return callback(
-          new ConversionError({
-            message: 'error converting file',
-            info: { bucket, originalKey, opts }
-          }).withCause(err)
-        )
+  let promise
```
gh2k: We could simplify this further if this was `let destPath` and then we followed up with `destpath = await blah` in the conditional section. We'd then only need one `try`/`catch` block.

package.json
```diff
@@ -21,7 +21,6 @@
   },
   "dependencies": {
     "@overleaf/o-error": "^2.1.0",
-    "async": "~0.2.10",
```
gh2k: :tada:

app/js/FileHandler.js
das7pad: I chose to split this in regards to line 124 following -- we would rethrow this error with a different message from the catch block then.

gh2k: OK, fair enough. :slightly_smiling_face:

das7pad:
Rebased to resolve conflicts.

PR #99, opened 2020-02-28
run npm update

briangough:
### Description

Update npm packages with `npm update` and `npm install`

Notable update aws-sdk 2.315.0 -> 2.628.0

#### Screenshots

NA

#### Related Issues / PRs

?

### Review

No changes to package.json, just updates with compatible versions.

https://github.com/aws/aws-sdk-js/blob/master/CHANGELOG.md#23810
2.381.0 bugfix: S3: fix putObject using stream <1mb
https://github.com/aws/aws-sdk-js/blob/master/CHANGELOG.md#23680
2.368.0 bugfix: md5: enable SDK to calculate content MD5 for more new S3 operations

#### Potential Impact

Low

#### Manual Testing Performed

- [ ] Unit and acceptance tests 

#### Accessibility

NA

### Deployment

NA

#### Deployment Checklist

NA

#### Metrics and Monitoring

NA

#### Who Needs to Know?

 @gh2k  here is the change from running npm update.

PR #100, opened 2020-03-02
Fix prettier task

gh2k:
# Description

The latest version of Prettier wants to be run with a full path instead of a relative one. This change calls Prettier with `$PWD` prepended to the path.

#### Related Issues / PRs

Introduced in #99 

### Review

Fix from https://stackoverflow.com/questions/60103564/there-was-trouble-creating-the-eslint-cliengine


gh2k:
I'm not sure it really matters. I thought we'd enabled `no-console` everywhere, but obviously not. I guess there's a service or two that it doesn't work for.

Seems more sensible to leave it out than have a hack that means we have to make the build scripts more complicated.

PR #104, opened 2020-03-19
Fix fallback to custom s3 buckets

gh2k:
### Description

Falling back to S3 custom buckets (e.g. for latexqc) was not working, as the fallback bucket would be null if not defined in the settings.

We should pass through the bucket instead of returning nothing, if we don't have a fallback for it. The worst that could happen is a 4xx error.

#### Review

I've tested this on staging. Properly this time.

I made the change to the tests first, and they failed until I added the fix to `MigrationPersistor.js`

### DIFF
--- a/app/js/MigrationPersistor.js
+++ b/app/js/MigrationPersistor.js
@@ -115,7 +115,7 @@ module.exports = function(primary, fallback) {
   }
 
   function _getFallbackBucket(bucket) {
-    return Settings.filestore.fallback.buckets[bucket]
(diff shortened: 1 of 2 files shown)


gh2k:
Yes. `copyOnMiss` is enabled in staging. It fails silently and logs a warning: https://github.com/overleaf/filestore/blob/62472e613d7e7be3a8a34c205eb7295ed77aa0a9/app/js/MigrationPersistor.js#L142

It copies the file in the background so doesn't affect the download.

PR #106, opened 2020-03-25
Use single pipeline when calculating md5

gh2k:
### Description

Adds a new `Stream.Transform` class that calculates the hash *and* byte count of a stream and passes it through. This removes the need for a separate metered stream module.

The idea here is that we have one single pipeline, instead if splitting the incoming data into two.

### Review

I've left in the `calculateStreamMd5` method, as this is called from the MD5-calculation endpoint and is useful when we want an MD5 but don't want to pass the stream through to anything else.

The dependency on `stream-meter` has been removed.

I've removed a couple of unit tests about egress/ingress metrics, as these were faffy to implement. I think there is very good acceptance-test coverage here, so I don't think that this is a problem.

#### Potential Impact

Most things.

### DIFF
--- a/app/js/PersistorHelper.js
+++ b/app/js/PersistorHelper.js
@@ -1,17 +1,54 @@
 const crypto = require('crypto')
-const metrics = require('metrics-sharelatex')
-const meter = require('stream-meter')
 const Stream = require('stream')
 const logger = require('logger-sharelatex')
+const metrics = require('metrics-sharelatex')
 const { WriteError, ReadError, NotFoundError } = require('./Errors')
 const { promisify } = require('util')
 
 const pipeline = promisify(Stream.pipeline)
 
+// Observes data that passes through and computes some metadata for it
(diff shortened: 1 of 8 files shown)


app/js/PersistorHelper.js
```diff
@@ -1,24 +1,62 @@
...
+// - if 'metric' is supplied as an option, this metric will be incremented by
+//   the number of bytes transferred
+class ObserverStream extends Stream.Transform {
+  constructor(options) {
+    super(options)
+
+    this.bytes = 0
+
+    if (options.hash) {
+      this.hash = crypto.createHash(options.hash)
```
das7pad: I am looking forward to seeing a stronger hash function used here. :+1:

gh2k: Google only give us MD5. It's fast and cheap, and works well enough for verifying integrity. Hash collisions aren't likely to be a problem for that scenario.

PR #107, opened 2020-03-26
(Hopefully) fix handle leaks

gh2k:
### Description

Try to fix handle leaks.

The most significant thing here was adding https://github.com/overleaf/filestore/commit/04c41816fe690155f325d9024ee11516533b4326 - which fails acceptance tests on unhandled promise rejection. This pointed out unhandled errors from the `pipeline` method when waiting for streams to become readable, which are fixed in https://github.com/overleaf/filestore/commit/d6170c95ee6e2520ad834b133d48cf45a022d9ae

For belt and braces, we also use autodestroy on the transform stream: https://github.com/overleaf/filestore/commit/cab1e3c21e0bc1129f52fdafc58914cf3b352cd1
And pass errors to the `destroy` method in the error handler: https://github.com/overleaf/filestore/commit/f4d08b73c195042cf93f7ac791c3a0932646c1ae

I don't know if either of these will make a difference, but they can't hurt.

### Review

This exposes a subtle misunderstanding in my knowledge of `pipeline`. Specifically, I had a mental model of it looking like this:

```
stream A  === data and errors ===> stream B === data and errors ===> stream C === errors ===> caught by pipeline
```
where it really works like
```
stream A === data ===> stream B === data ===> stream C
   |                      |                      |
   \/                    \/                     \/
pipeline                pipeline              pipeline
```

meaning that you can't expect the last stream in the pipe to emit an error event just because an earlier one does... You need to listen to the `pipeline` itself.

### DIFF
--- a/app/js/PersistorHelper.js
+++ b/app/js/PersistorHelper.js
@@ -50,7 +50,7 @@ module.exports = {
   ObserverStream,
   calculateStreamMd5,
   verifyMd5,
-  waitForStreamReady,
+  getReadyPipeline,
   wrapError,
   hexToBase64,
   base64ToHex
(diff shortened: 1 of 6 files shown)


app/js/PersistorHelper.js
```diff
@@ -23,8 +24,11 @@ class ObserverStream extends Stream.Transform {
       this.hash = crypto.createHash(options.hash)
     }
     if (options.metric) {
-      const onEnd = () => {
+      const onEnd = err => {
         metrics.count(options.metric, this.bytes)
+        if (err) {
+          this.destroy(err)
+        }
```
das7pad: The `.destroy(err)` invocation will emit another `'error'` event, which is not allowed per the docs.

> https://nodejs.org/api/stream.html#stream_writable_destroy_error
> 
> `writable.destroy([error])`
> - error `<Error>` Optional, an error to emit with `'error'` event.


> https://nodejs.org/api/stream.html#stream_event_error
> 
> After `'error'`, no further events other than `'close'` should be emitted (including `'error'` events).


---

Using just `.autoDestroy = true` should be fine.

> https://nodejs.org/api/stream.html#stream_event_error
> 
> The stream is not closed when the 'error' event is emitted unless the autoDestroy option was set to true when creating the stream.

gh2k: :+1: thanks

gh2k:
force-pushed to remove the destroy commit

PR #108, opened 2020-03-27
Retry file reads on 5xx error from upstream

gh2k:
### Description

We've noticed a number of `502` and `504` errors in the logs - particularly from `filestore-readonly`, which fails the download immediately, possibly breaking a compile in the process.

This patch causes us to retry a few times if we get these errors. They seem very transient so I suspect that retrying will help.

### Review

This only affects `getFoo`, and only when the download fails the first time and we would have thrown an error anyway... So I think it's fairly safe to try.

I couldn't make an acceptance test for this against the fake GCS server, so the unit tests will have to do.

#### Potential Impact

Downloading files, when errors occur

### DIFF
--- a/app/js/FileHandler.js
+++ b/app/js/FileHandler.js
@@ -1,11 +1,19 @@
-const { callbackify } = require('util')
+const { callbackify, promisify } = require('util')
 const fs = require('fs')
+const OError = require('@overleaf/o-error')
+const Settings = require('settings-sharelatex')
 const PersistorManager = require('./PersistorManager')
 const LocalFileWriter = require('./LocalFileWriter')
 const FileConverter = require('./FileConverter')
 const KeyBuilder = require('./KeyBuilder')
 const ImageOptimiser = require('./ImageOptimiser')
-const { ConversionError, InvalidParametersError } = require('./Errors')
+const {
+  ConversionError,
+  InvalidParametersError,
+  UpstreamError
+} = require('./Errors')
+
+const sleep = promisify(setTimeout)
 
 module.exports = {
   insertFile: callbackify(insertFile),
@@ -63,18 +71,27 @@ async function deleteProject(bucket, key) {
 async function getFile(bucket, key, opts) {
   opts = opts || {}
   if (!opts.format && !opts.style) {
-    return PersistorManager.promises.getFileStream(bucket, key, opts)
+    return _wrapWithRetry(
+      PersistorManager.promises.getFileStream,
+      bucket,
+      key,
+      opts
+    )
   } else {
-    return _getConvertedFile(bucket, key, opts)
+    return _wrapWithRetry(_getConvertedFile, bucket, key, opts)
   }
 }
 
 async function getFileSize(bucket, key) {
-  return PersistorManager.promises.getFileSize(bucket, key)
+  return _wrapWithRetry(PersistorManager.promises.getFileSize, bucket, key)
 }
 
 async function getDirectorySize(bucket, projectId) {
-  return PersistorManager.promises.directorySize(bucket, projectId)
+  return _wrapWithRetry(
+    PersistorManager.promises.directorySize,
+    bucket,
+    projectId
+  )
 }
 
 async function _getConvertedFile(bucket, key, opts) {
@@ -174,4 +191,25 @@ async function _writeFileToDisk(bucket, key, opts) {
     opts
   )
   return LocalFileWriter.promises.writeStream(fileStream, key)
+}
+
+async function _wrapWithRetry(method, ...params) {
+  const maxTries = Settings.maxRetries || 5
+  let delay = Settings.retryDelay || 125
+  let tries = 0
+
+  while (true) {
+    try {
+      const result = await method(...params)
+      return result
+    } catch (err) {
+      if (OError.hasCauseInstanceOf(err, UpstreamError) && tries < maxTries) {
+        await sleep(delay)
+        tries++
+        delay *= 2
+      } else {
+        throw err
+      }
+    }
+  }
 }
--- a/app/js/FileHandler.js
+++ b/app/js/FileHandler.js
@@ -204,7 +204,7 @@ async function _wrapWithRetry(method, ...params) {
       return result
     } catch (err) {
       if (OError.hasCauseInstanceOf(err, UpstreamError) && tries < maxTries) {
-        await sleep(delay)
+        await sleep(delay * (Math.random() + 0.5))
         tries++
         delay *= 2
       } else {
--- a/app/js/PersistorHelper.js
+++ b/app/js/PersistorHelper.js
@@ -2,7 +2,12 @@ const crypto = require('crypto')
 const Stream = require('stream')
 const logger = require('logger-sharelatex')
 const metrics = require('metrics-sharelatex')
-const { WriteError, ReadError, NotFoundError } = require('./Errors')
+const {
+  WriteError,
+  ReadError,
+  NotFoundError,
+  UpstreamError
+} = require('./Errors')
 const { promisify } = require('util')
 
 const pipeline = promisify(Stream.pipeline)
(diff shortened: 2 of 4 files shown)


app/js/FileHandler.js
```diff
@@ -175,3 +192,24 @@ async function _writeFileToDisk(bucket, key, opts) {
...
+  let delay = Settings.retryDelay || 125
+  let tries = 0
+
+  while (true) {
+    try {
+      const result = await method(...params)
+      return result
+    } catch (err) {
+      if (OError.hasCauseInstanceOf(err, UpstreamError) && tries < maxTries) {
+        await sleep(delay)
```
das7pad: Let's add a random delay here. Otherwise we will hit the backend in large batches on second try and overload it again.

app/js/PersistorHelper.js
```diff
@@ -131,6 +136,15 @@ function wrapError(error, message, params, ErrorType) {
       message: 'no such file',
       info: params
     }).withCause(error)
+  } else if (
+    error instanceof UpstreamError ||
+    parseInt(error.code) >= 500 ||
+    (error.message && error.message.match(/^Cannot parse response as JSON.*/))
+  ) {
+    return new UpstreamError({
+      message: 'internal error from GCS',
```
das7pad: So this actually `GCSUpstreamError`?

gh2k: No, I should make the error more generic. (If S3 returns a 5xx error code, this logic will kick in as well)

das7pad: > (If S3 returns a 5xx error code, this logic will kick in as well)

Can you adjust the message then, to not include `GCS`?

gh2k: Yeah, that's what I meant :-)

das7pad: I should read more closely :D

jdleesmiller:
Just to check: can we use the built-in retries in the GCS client? It looks like there is some retry logic.

https://github.com/googleapis/nodejs-common/blob/9f410477432893f68e57b5eeb31a068a3d8ef52f/src/util.ts#L756

das7pad:
> Just to check: can we use the built-in retries in the GCS client? It looks like there is some retry logic.
> 
> https://github.com/googleapis/nodejs-common/blob/9f410477432893f68e57b5eeb31a068a3d8ef52f/src/util.ts#L756

It looks like it already does retry?

https://github.com/googleapis/nodejs-common/blob/9f410477432893f68e57b5eeb31a068a3d8ef52f/src/util.ts#L314-L318
> <pre> 
>  /**
>   * Maximum number of automatic retries attempted before returning the error.
>   * (default: 3)
>   */
> </pre>
>  maxRetries?: number;

jdleesmiller:
> It looks like it already does retry?

Yes, that is my understanding, at least for GETs. It doesn't retry on 504, but it does retry on 502, it seems:

https://github.com/googleapis/nodejs-common/blob/9f410477432893f68e57b5eeb31a068a3d8ef52f/src/util.ts#L517

I guess we could try bumping the number of retries in the config...

https://github.com/googleapis/nodejs-storage/blob/39869e3c6c62eabe1840f0fd884b361265e2cb76/src/storage.ts#L48-L57

but really 3 should be enough.

Under the hood, it looks like it uses https://www.npmjs.com/package/retry-request, so there may be some traces we could pick up to see if the request is being retried internally or not. (Or we could try enabling the `DEBUG` for the request-retry library.)

gh2k:
> built-in retries

Hmm, yes it does have that.

It uses `retry-request`, which also has a backoff:
https://github.com/stephenplusplus/retry-request/blob/6f69c95dceada6b4255c148fbd27c4d530fcec3e/index.js#L213-L215

I wonder if any of:
1. we need more retries
1. the gcs client is not retrying, or
1. we need a longer delay

...are true

~The `502` errors are much more common than the `504`s, and the error here usually presents as `https://github.com/overleaf/filestore/blob/f7deb6be27886446a51ab56ee409e401ec33b24d/app/js/PersistorHelper.js#L142` with an HTML response body in the error.~

Upon further inspection, the 502s are coming only from `POST` requests in filestore, so there's nothing we can easily do about that.

* [Example of a 502 error](https://console.cloud.google.com/logs/viewer?project=mgcp-1117973-ol-prod&folder&organizationId&minLogLevel=0&expandAll=false&timestamp=2020-03-27T12%3A37%3A35.549000000Z&customFacets&limitCustomFacetWidth=true&dateRangeStart=2020-03-26T12%3A37%3A35.800Z&dateRangeEnd=2020-03-27T12%3A37%3A35.800Z&interval=P1D&resource=k8s_container%2Fcluster_name%2Fol-gke-02%2Fnamespace_name%2Fdefault%2Fcontainer_name%2Ffilestore&scrollTimestamp=2020-03-27T11%3A45%3A31.094000000Z&filters=text%3A%22Cannot%20parse%20response%20as%20JSON%22&advancedFilter=resource.type%3D%22k8s_container%22%0Aresource.labels.container_name%3D%22filestore%22%0Aresource.labels.namespace_name%3D%22default%22%0Aresource.labels.location%3D%22us-east1%22%0Aresource.labels.project_id%3D%22mgcp-1117973-ol-prod%22%0Aresource.labels.pod_name%3D%22filestore-prod-59cccdb69f-hjtjp%22%0Aresource.labels.cluster_name%3D%22ol-gke-02%22%0Atimestamp%3D%222020-03-27T06%3A34%3A22.226000000Z%22%0AinsertId%3D%22tmizoj684c4vpr1lv%22)
* [Example of a 504 error](https://console.cloud.google.com/logs/viewer?project=mgcp-1117973-ol-prod&folder&organizationId&minLogLevel=0&expandAll=false&timestamp=2020-03-27T12%3A34%3A22.119000000Z&customFacets&limitCustomFacetWidth=true&dateRangeStart=2020-03-27T06%3A34%3A22.370Z&dateRangeEnd=2020-03-27T12%3A34%3A22.370Z&interval=PT6H&resource=k8s_container%2Fcluster_name%2Fol-gke-01%2Fnamespace_name%2Fdefault%2Fcontainer_name%2Ffilestore-readonly&scrollTimestamp=2020-03-27T11%3A44%3A32.492000000Z&filters=text%3A502&advancedFilter=resource.type%3D%22k8s_container%22%0Aresource.labels.project_id%3D%22mgcp-1117973-ol-prod%22%0Aresource.labels.cluster_name%3D%22ol-gke-01%22%0Aresource.labels.pod_name%3D%22filestore-readonly-prod-766db95cf7-6ddk4%22%0Aresource.labels.container_name%3D%22filestore-readonly%22%0Aresource.labels.namespace_name%3D%22default%22%0Aresource.labels.location%3D%22us-east1-b%22%0Atimestamp%3D%222020-03-27T09%3A25%3A53.829000000Z%22%0AinsertId%3D%22c40rfk2njmearg0uq%22)

My initial thought was that `retry-request` might not be retrying on 504 in that case, but that doesn't seem to be it: https://github.com/stephenplusplus/retry-request/blob/6f69c95dceada6b4255c148fbd27c4d530fcec3e/index.js#L22

504 doesn't happen very often, so I'm happy to drop this if it's just for that case. Betwen 1pm and 3pm yesterday it happened 4 times.

One thing that occurs though: If the `502` gets sent back from the initial POST request before we write any data to it, then we could theoretically do something similar to the wait-for-stream-to-become-ready mechanism that we have for reads, which waits to see if we're going to get a 404 before sending the stream back to the client. I don't know whether Google will do that or let us merrily send the whole of the POST before it gives up.

There's no `writeable` event, but we could maybe `await write('')` and try again if we can't get that far?

... or, we could just get the client to try again. :man_shrugging: 

For the 502s, we've been seeing about 1-2 an hour as well, so maybe it's not worth it for this either, seeing as it's something of a speculative fix. I'd be happy to close this. tbh I thought they were happening more often.

jdleesmiller:
> the 502s are coming only from POST requests in filestore

OK, I think that makes sense. It won't be retrying streaming POST requests, looking at the code:
https://github.com/googleapis/nodejs-common/blob/9f410477432893f68e57b5eeb31a068a3d8ef52f/src/util.ts#L758-L762

So... I guess the question is whether we want to add our own retry for the POST case. Is that a simpler task?

(Or we could avoid streaming for the POST case, in which case it would retry, but then we might have other problems with large files.)

gh2k:
I'd rather not get the disk involved if I can help it. For one, it means buffering the entire file before sending it on, which will slow things down.

I don't think we can retry a post once it's started, as if we read from the stream then we can't start again, which is why they're hard. Possibly "faff with the connection for a bit and see if it errors before sending any data" might work, but it might not. For something this infrequent I think it's best to leave it. Will close.

PR #109, opened 2020-03-30
Call end() when handing ERR_STREAM_PREMATURE_CLOSE

gh2k:
### Description

It looks like we are not doing anything on an `ERR_STREAM_PREMATURE_CLOSE` error. I think this was an attempt to not write errors to the logs when this happens.

On other errors, we are passing on to `next`. On no error, I would assume that the request has completed successfully. This patch tries to close the connection when a premature-close error occurs, rather than leaving it hanging.

See further diagnosis at: https://github.com/overleaf/issues/issues/2880#issuecomment-606033804

### DIFF
--- a/app/js/FileController.js
+++ b/app/js/FileController.js
@@ -61,18 +61,18 @@ function getFile(req, res, next) {
     }
 
     pipeline(fileStream, res, err => {
-      if (err) {
+      if (!fileStream.destroyed) {
         fileStream.destroy()
-        if (err.code === 'ERR_STREAM_PREMATURE_CLOSE') {
-          res.end()
-        } else {
-          next(
-            new Errors.ReadError({
-              message: 'error transferring stream',
-              info: { bucket, key, format, style }
-            }).withCause(err)
-          )
-        }
+      }
+      if (err && err.code === 'ERR_STREAM_PREMATURE_CLOSE') {
+        res.end()
+      } else if (err) {
+        next(
+          new Errors.ReadError({
+            message: 'error transferring stream',
+            info: { bucket, key, format, style }
+          }).withCause(err)
+        )
       }
     })
   })
--- a/app/js/FileController.js
+++ b/app/js/FileController.js
@@ -61,7 +61,9 @@ function getFile(req, res, next) {
     }
 
     pipeline(fileStream, res, err => {
-      if (err && err.code !== 'ERR_STREAM_PREMATURE_CLOSE') {
+      if (err && err.code === 'ERR_STREAM_PREMATURE_CLOSE') {
+        res.end()
+      } else if (err) {
         next(
           new Errors.ReadError({
             message: 'error transferring stream',
--- a/app/js/FileController.js
+++ b/app/js/FileController.js
@@ -61,15 +61,18 @@ function getFile(req, res, next) {
     }
 
     pipeline(fileStream, res, err => {
-      if (err && err.code === 'ERR_STREAM_PREMATURE_CLOSE') {
-        res.end()
-      } else if (err) {
-        next(
-          new Errors.ReadError({
-            message: 'error transferring stream',
-            info: { bucket, key, format, style }
-          }).withCause(err)
-        )
+      if (err) {
+        fileStream.destroy()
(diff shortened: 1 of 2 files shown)


gh2k:
I think you may be right... That was going to be the next thing I looked at tbh.

In terms of things that are obviously wrong, the thing that this patch addresses looks likely it ought to be an error, as we are deliberately squelching an error response from `pipeline`.

I was planning to add a `writeStream.destroy` into the error handler (or even outside it) that is modified here if this doesn't work, but I am wary of double-destroying it.

`fileStream` in this case is an `ObserverStream`, which has `autoDestroy` set to `true`. There will be an original `pipeline` command hanging around though from https://github.com/overleaf/filestore/blob/746430eb469a5f870372484f941f10f617ab743f/app/js/PersistorHelper.js#L117 which may not have seen an error at either end. That means that we don't immediately have access to the original stream.

That `pipeline` call has a catch handler, which might also be an issue.. Although if it throws an error at that point we'll wind up with an unhandled promise rejection rather than finding out about it in the controller.

So tbh I don't know whether at this point:
1. Whether we have the `ObserverStream` still knocking around, or if that goes away when the pipeline errors
1. Whether getting rid of the `res` will somehow let the `ObserverStream` know that it's OK to destroy
1. Whether destroying the `ObserverStream` will destroy the GCS stream
1. Whether an error is getting emitted by the `ObserverStream` and caught by the catch handler for the initial pipeline

Clearly the GCS stream *usually* gets destroyed, or we'd be running out of handles very quickly.

Belt & braces would be:
1. Call `destroy` on the `fileStream` in the `FileController` error handler
1. Call `destroy` all of `...streams` in the `PersistorHandler` catch handler

I guess that double-destroying might not be an issue here as we're already in an error state, provided we don't crash filestore in the process?

gh2k:
I think I may be thinking with my `C` head again...

```JavaScript
const fs = require('fs')
const Stream = require('stream')

const stream = fs.createReadStream('file.txt')

for (const e of ['error', 'close', 'end']) {
  stream.on(e, x => console.log(e, x))
}

console.log('destroying the stream')
stream.destroy()

console.log('destroying the stream again')
stream.destroy()

const out = fs.createWriteStream('/dev/null')
console.log('doing something with the stream')
Stream.pipeline(stream, out, err => {
  console.log('pipeline', err.message)
  process.exit(0)
})

```
says
```
destroying the stream
destroying the stream again
doing something with the stream
close undefined
pipeline Premature close
```
so I *think* it's probably OK...

briangough:
Based on this old commit, I think it's safe to call destroy twice
https://github.com/nodejs/node-v0.x-archive/commit/03eb41c2ec4e99c52a5338ecd021056d065ddaab

das7pad:
> Based on this old commit, I think it's safe to call destroy twice

The logic has moved, but remains the same.

https://github.com/nodejs/node/blob/v12.16.1/lib/_stream_writable.js#L770
https://github.com/nodejs/node/blob/v12.16.1/lib/internal/streams/destroy.js#L11-L23

app/js/FileController.js
```diff
@@ -61,18 +61,18 @@ function getFile(req, res, next) {
     }
 
     pipeline(fileStream, res, err => {
-      if (err) {
+      if (!fileStream.destroyed) {
```
das7pad: This was not moved out of the `if (err ...)` branch, was that intentional?

gh2k: Yes. I figured that if we can be sure not to destroy things unnecessarily we may as well check it even if there hasn't been an error.

briangough:
Just writing up my suspicions that it's to do with calling destroy on a transform stream.  It says that the default implementation will emit `close` - perhaps that triggers this error.  It's not clear whether it will pass it through to the underlying stream.

```
##### transform.destroy([error])
<!-- YAML
added: v8.0.0
-->
* `error` {Error}

Destroy the stream, and emit `'error'`. After this call, the
transform stream would release any internal resources.
Implementors should not override this method, but instead implement
[`readable._destroy()`][readable-_destroy].
The default implementation of `_destroy()` for `Transform` also emit `'close'`.
```

PR #111, opened 2020-03-31
Only delete the converted-cache folder if conversions are enabled

gh2k:
### Description

Don't call `deleteDirectory` on the `converted-cache` folder when inserting or deleting files, unless conversions are enabled.

Slack discussion: https://digital-science.slack.com/archives/G6256GXGS/p1585663651303300

#### Review

It's highly probable that we don't need to delete the directory on `insertFile` based on how we expect uploads to work, but I would rather change as little as possible so I'm leaving this in place for now.

### DIFF
--- a/app/js/FileHandler.js
+++ b/app/js/FileHandler.js
@@ -51,7 +51,7 @@ async function deleteFile(bucket, key) {
   if (Settings.enableConversions) {
     jobs.push(PersistorManager.promises.deleteDirectory(bucket, convertedKey))
   }
-  await Promise.all([jobs])
+  await Promise.all(jobs)
 }
 
 async function deleteProject(bucket, key) {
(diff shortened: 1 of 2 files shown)


app/js/FileHandler.js
```diff
@@ -44,10 +47,11 @@ async function deleteFile(bucket, key) {
...
   }
-  await Promise.all([
-    PersistorManager.promises.deleteFile(bucket, key),
-    PersistorManager.promises.deleteDirectory(bucket, convertedKey)
-  ])
+  const jobs = [PersistorManager.promises.deleteFile(bucket, key)]
+  if (Settings.enableConversions) {
+    jobs.push(PersistorManager.promises.deleteDirectory(bucket, convertedKey))
+  }
+  await Promise.all([jobs])
```
das7pad:
```suggestion
  await Promise.all(jobs)
```

PR #112, opened 2020-03-31
Destroy download streams on close

gh2k:
### Description

This modifies the `getReadyPipeline` helper to do two things:

* Ensures that streams are destroyed on error, when the error is caught before the promise rejects
* Calls `destroy` on streams in the pipeline when the next one emits `close`

### Review

I realise that this is throwing stuff at the wall to see what sticks somewhat, but here we are.

### DIFF
--- a/app/js/PersistorHelper.js
+++ b/app/js/PersistorHelper.js
@@ -106,11 +106,12 @@ function getReadyPipeline(...streams) {
 
         lastStream.removeListener('readable', handler)
         if (err) {
-          return reject(
+          reject(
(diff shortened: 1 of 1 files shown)


app/js/PersistorHelper.js
```diff
@@ -121,6 +122,9 @@ function getReadyPipeline(...streams) {
       }
     }
 
+    for (let index = 0; index < streams.length - 1; index++) {
```
das7pad: Totally not `c` :)

PR #114, opened 2020-04-03
Add fast-crc32c library

gh2k:
### Description

Google recommend installing `fast-crc32c` to improve the performance of crc computation:
https://googleapis.dev/nodejs/storage/latest/File.html#createWriteStream

It's possible that this will help reduce resource usage if we add it.


gh2k:
> This is not the correct library
facepalm8

PR #115, opened 2020-04-03
Fix error handler returning two responses

emcsween:
### Description

The error handler mistakenly sent two responses on error, the first
being a 200.

#### Related Issues / PRs

https://github.com/overleaf/issues/issues/2925

#### Potential Impact

Everything related to file storage.

#### Manual Testing Performed

- [x] Upload a file with GCS turned off. Upload fails

### DIFF
--- a/app/js/RequestLogger.js
+++ b/app/js/RequestLogger.js
@@ -17,10 +17,7 @@ class RequestLogger {
 
   static errorHandler(err, req, res, next) {
     req.requestLogger.addFields({ error: err })
-    res
-      .send(err.message)
-      .status(500)
-      .end()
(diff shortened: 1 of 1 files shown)


40thieves:
Is this a problem that could be caught by linting?

emcsween:
I suppose we could write a linting rule that catches this kind of pattern where a sending method is called twice on the response, but it would have to rely on the response being called `res` and on the name of the calls (`send()`, `end()`). It wouldn't work if you send `res` to another function that does the calls... I don't think it's worth the effort, TBH.

40thieves:
Fair enough 👍

PR #116, opened 2020-04-15
Add support for redirecting to signed URLs

gh2k:
### Description

Add a mechanism to retrieve a signed URL from filestore and redirect to it, instead of proxying the file directly.

Falls back to proxying if the url can't be retrieved.

#### Related Issues / PRs

Fixes overleaf/issues#2937

### Review

I've only added support for redirects to the GCS persistor, as this is where the bulk of the traffic goes. If we're using the S3 persistor now it's because we're falling back to an unusual bucket.

It's running in staging with redirect URLs only enabled for `filestore-readonly`. I think this is sensible for now as that's where the bulk of the load comes from. Compiles still work and I'm seeing hits on the `file_redirect` metric.

I couldn't get acceptance tests working against this, or signed URLs working in dev. Tested in staging.

#### Potential Impact

Requires a settings change to proxy the files - assuming we're just making this change for readonly, this affects compiles.

#### Manual Testing Performed

- [x] Tested compiles in staging with this setting turned on

### Deployment

Requires setting `ALLOW_REDIRECTS: 'true'` for `filestore-readonly`

#### Metrics and Monitoring

There are two new metrics: `file_redirect` and `file_redirect_error` so that we can watch for problems and make sure that it's doing what we want.

### DIFF
--- a/app/js/FileHandler.js
+++ b/app/js/FileHandler.js
@@ -13,10 +13,12 @@ module.exports = {
   deleteFile: callbackify(deleteFile),
   deleteProject: callbackify(deleteProject),
   getFile: callbackify(getFile),
+  getRedirectUrl: callbackify(getRedirectUrl),
   getFileSize: callbackify(getFileSize),
   getDirectorySize: callbackify(getDirectorySize),
   promises: {
     getFile,
+    getRedirectUrl,
     insertFile,
     deleteFile,
     deleteProject,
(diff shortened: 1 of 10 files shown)


app/js/GcsPersistor.js
```diff
@@ -141,6 +143,26 @@ async function getFileStream(bucketName, key, _opts = {}) {
...
 }
 
+async function getRedirectUrl(bucketName, key) {
+  try {
+    const [url] = await storage
+      .bucket(bucketName)
+      .file(key)
+      .getSignedUrl({
+        action: 'read',
+        expires: new Date().getTime() + settings.filestore.signedUrlExpiryInMs
```
briangough, line 153: `Date.now() + settings.filestore.signedUrlExpiryInMs` also works.

app/js/FileHandler.js
```diff
@@ -73,6 +75,24 @@ async function getFile(bucket, key, opts) {
...
+  // if we're doing anything unusual with options, or the request isn't for
+  // one of the default buckets, return null so that we proxy the file
+  opts = opts || {}
+  if (
+    !opts.start &&
+    !opts.end &&
+    !opts.format &&
+    !opts.style &&
+    Object.values(Settings.filestore.stores).includes(bucket) &&
+    Settings.filestore.allowRedirects
```
briangough, line 88: Maybe check for opts being completely empty (Object.keys(opts).length === 0)? I think we can only support redirects for that case.

gh2k: I started with that, but changed it to this way. I think `opts` comes populated with some redundant things like the bucket name, possibly from some of the route-mangling middleware. So we'd then be in the business of stripping out the stuff we don't need, which would then just end up being some inverse of this. I thought it better to just test for these things.

briangough: Ah fair enough, I hadn't realised there was other stuff in there.  Looks like it is not actually used though.

PR #117, opened 2020-05-05
upgrade to o-error v3

warncke:
### Description

Upgrade to OError v3 and change signature on error instantiations.

#### Related Issues / PRs

https://github.com/overleaf/issues/issues/3214

#### Potential Impact

???

#### Manual Testing Performed

- [ ] ???

### DIFF
--- a/app/js/LocalFileWriter.js
+++ b/app/js/LocalFileWriter.js
@@ -30,9 +30,9 @@ async function writeStream(stream, key) {
   } catch (err) {
     await deleteFile(fsPath)
 
-    throw new WriteError({
-      message: 'problem writing file locally',
-      info: { err, fsPath }
+    throw new WriteError('problem writing file locally', {
+      err,
+      fsPath
     }).withCause(err)
   }
 }
@@ -45,10 +45,7 @@ async function deleteFile(fsPath) {
     await promisify(fs.unlink)(fsPath)
   } catch (err) {
     if (err.code !== 'ENOENT') {
-      throw new WriteError({
-        message: 'failed to delete file',
-        info: { fsPath }
-      }).withCause(err)
+      throw new WriteError('failed to delete file', { fsPath }).withCause(err)
     }
   }
 }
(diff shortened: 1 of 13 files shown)


app/js/LocalFileWriter.js
```diff
@@ -30,9 +30,9 @@ async function writeStream(stream, key) {
   } catch (err) {
     await deleteFile(fsPath)
 
-    throw new WriteError({
-      message: 'problem writing file locally',
-      info: { err, fsPath }
+    throw new WriteError('problem writing file locally', {
+      err,
```
jdleesmiller, line 34: Note: I think this probably shouldn't have been in the `info`.

package.json
```diff
@@ -21,7 +21,7 @@
   },
   "dependencies": {
     "@google-cloud/storage": "^4.7.0",
-    "@overleaf/o-error": "^2.1.0",
+    "@overleaf/o-error": "git://github.com/overleaf/o-error.git#14e515a195d6dbd3711c5a211a730752802d3b03",
```
40thieves, line 24: Just flagging that we'll have to swap this with the final 3.x release.

mserranom:
Hey @warncke, what's the status of this PR?

mserranom:
It's probably worth updating to the latest `logger-sharelatex` as part of the PR.

test/unit/js/GcsPersistorTests.js
```diff
@@ -44,6 +44,7 @@ describe('GcsPersistorTests', function() {
           user_files: 'user_files'
         },
         gcs: {
+          deleteConcurrency: 1,
```
gh2k, line 47: Did this need to go in here? It will disable concurrency on the delete-directory test.

gh2k:
> I agree that updating logger-sharelatex is worth doing here or soon after as follow up

If this is not being done here, could we open a ticket to make sure it gets done? ty

jdleesmiller:
I think this is back on @warncke when he's back but should be ready to deploy very soon.

test/unit/js/GcsPersistorTests.js
warncke: i added this because of some error that was coming up in unit testing, but it must have been caused by something else I fixed, because the tests passed once I took this back out.

warncke:
Upgrade logger-sharelatex but did not redo the withCause to tag

PR #136, opened 2020-10-13
Upgrade object-persistor to latest version

gh2k:
### Description

Deleting an empty directory can cause problems in the dev environment. There is a fix here:
https://github.com/overleaf/object-persistor/pull/12

This patch pulls in that change.

#### Related Issues / PRs

Fix: https://github.com/overleaf/object-persistor/pull/12
Should unblock: https://github.com/overleaf/web-internal/pull/3269

### Review

Just a package-lock upgrade. I've checked the commit hash against the head of master in object-persistor.

#### Potential Impact

Should be fairly limited.


gh2k:
Force-pushed to unmangle double-commit on package-lock

briangough (approved):

